Issue #236, opened 2022-04-20
Merge Meltano and Custom taps/targets


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/hub/-/issues/240

Originally created by @pnadolny13 on 2022-04-20 21:01:53

---

A note from https://gitlab.com/meltano/hub/-/merge_requests/237:

>Its important to note that for now since it was a little complex to blend them I decided that if there was a meltano optimized definition and a custom tap/target definition then I will just take the meltano one. My thought is that later we can combine the two files into a comprehensive single file.

If its not solved by https://gitlab.com/meltano/hub/-/issues/203 already we should make sure that we find a way to merge the Singer plugins listed in discovery.yml with the Singer plugins in the hub when they exist in both places. Right now we are just taking the discovery.yml definition if they both exist.

### Update 2022-05-27

From https://gitlab.com/meltano/hub/-/issues/240#note_962936968

The proposed plan to bring the current hub up to speed so it can do the job for meltano 2.0 before the redesign is complete:

* [x] refresh the hub meltano files https://gitlab.com/meltano/hub/-/tree/main/\_data/meltano based on any recent discovery.yml updates
* [x] implement the new blended yaml schema (currently only implemented in the hub redesign repo https://gitlab.com/meltano/hub-v2/-/tree/develop/data/plugins) in the current hub
* [x] do a one time manual update to any extractors/loaders that have missing variants due to the challenge of being added only to the custom defintion and not to discovery.yml
* [x] make any necessary updates to the API to serve the new files
* [x] make any necessary front end changes to use the new file format

This issue will primarily be used for the updates.

MeltyBot:
[View 12 previous comments from the original issue on GitLab](https://gitlab.com/meltano/hub/-/issues/240)

PR #514, opened 2022-05-31
P3 - replace data directory, add defaults map, remove tap/target logos

pnadolny13:
Relates to https://github.com/meltano/hub/issues/236

This is the part 1 PR that adds the changes to have a single file format for what were previously meltano and singer definitions. A second PR will follow to update all the code for API/UI to use the new format, it should be much easier to review the code since this PR has 1k+ files changing due to the definitions.

- Replaces the data directory with the new directory structure to split variants to independent files, it looks like: `_data/meltano/extractors/tap-csv/meltanolabs.yml`
- Remove the taps and targets definition files since they now live under extractors/loaders
- Removes dashboards/models definitions since they arent used by the UI or the API
- Removes taps and target logo directory, the extractors/loaders contain all necessary logos
- I did a manual diff to pull in all variants to solve the case where we were previously naively taking just the discovery.yml entry if it exists
- The count of extractors is 306 vs 307 on the published hub, this is because I consolidated meltanolabs tap-github under the tap-github umbrella instead of keeping it as a separate github search.


alexmarple:
@pnadolny13 I am running into this issue when trying to spin up this branch. I'm going to investigate further why this is happening. I'm wondering if this simply needs main to be merged into it to fix it.

Screen Shot 2022-06-01 at 11 39 03 AM

pnadolny13:
> I've added the footer fix and it all looks good to me!

@alexmarple Thanks for that!

Issue #253, opened 2022-05-27
Automate generation of the detail pages for Extractor / Loader Meltano plugins


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/hub/-/issues/260

Originally created by @aaronsteers on 2022-05-27 17:37:23

---

TODO:

1. [x] For testing purposes, create a dummy `.../loaders/target-snowflake/meltano-prereqs.md` and `.../target-snowflake/meltano-usage.md` files for meltano's default target-snowflake plugin. (Substitute another variant name if `meltanolabs` isn't the default.)
   - File text can be simple and dummy like "PREREQS GO HERE", "USAGE INFO GOES HERE".
   - These are adjacent to (and alpha sorted with) the yaml inputs such as `_data/meltano/loaders/target-snowflake/meltano.yml` 
1. [x] Add logic to the script that prepends the contents of a `*-prereqs.md` file if and appends the contents of a `*-usage.md` file it is present for any given plugin.
1. [x] Ensure the generated file lands in the right place.
1. [x] Resolve any naming issues if needed, just as distinction betweeen `tap-salesforce` and `salesforce`.

Removed (will be handled in other issues:

> 1. [x] Duplicate the markdown-driven page creation scripts that are used for generating tap and target detail pages, and apply that automated script to the Meltano plugin version of the detail pages. https://github.com/meltano/hub/pull/541
> 1. [x] Add logic to the script that puts a warning on the page if `settings` is missing, along with instructions and CTA for contributions. - https://github.com/meltano/hub/pull/528
> 1. [x] Add logic to the script that puts a warning on the page if `capabilities` is missing, along with instructions and CTA for contributions. - https://github.com/meltano/hub/pull/528
> 3. [x] Move https://docs.meltano.com/concepts/plugins#sqlfluff and https://docs.meltano.com/concepts/plugins#great-expectations to https://hub.meltano.com/utilities/sqlfluff and https://hub.meltano.com/utilities/great_expectations - https://github.com/meltano/hub/pull/531

MeltyBot:
[View 13 previous comments from the original issue on GitLab](https://gitlab.com/meltano/hub/-/issues/260)

DouweM:
@tayloramurphy @aaronsteers FYI I've updated the tasks based on my top/bottom proposal in https://gitlab.com/meltano/hub/-/issues/260#note_963069049.

DouweM:
[...]

PR #515, opened 2022-05-31
P1 - refactor API and UI to use the new definition directory structure

pnadolny13:
This includes the refactored API and UI code needed to use the new directory structure for plugin definitions.

Things to call out:

1. In this PR I'm preferring autogenerated files over the markdown files, I actually found that the generated ones look pretty good. So I'd prefer to use those to reduce maintenance and potentially stale data with the trade off of not having some of the more rich text and image/videos attached.
1. Removed the SingerHub tap/target pages and navigation
1. MeltanoHub landing page has a spot for maintainers now that links to the maintainers landing page
1. Fixed some docs links that were broken
1. The plugin API is updated and passes schema validation also
1. The Singer API is being removed here. It would be too hard for me to merge the new files back into the old multi variant format to support it. We can optionally check in the current state of those files into git then call it deprecated.
 

Updates:
- For note 2 above- Singer tap/targets pages will still be linked but they'll be identical to the extractor pages
- For 3 - the maintainers landing page will be generated but not linked from anywhere
- For 6  - Singer API files will be snapshotted and checked into the repo for backwards compatibility but will not be updated moving forward
- The count of extractors is 306 vs 307 on the published hub, this is because I consolidated meltanolabs tap-github under the tap-github umbrella instead of keeping it as a separate github search.


tayloramurphy:
@pnadolny13 couple of things I'm noticing:

* We've lost the "Meltano-optimized" icon - I'm not _super_ concerned about that as I think we should potentially auto-calculate that based on the settings / capabilities in https://github.com/meltano/hub/issues/253 
* The URLs once you click into extractors have `/taps` and not `/extractors` (same for loaders / targets)

pnadolny13:
> @pnadolny13 couple of things I'm noticing:
> 
> * We've lost the "Meltano-optimized" icon - I'm not _super_ concerned about that as I think we should potentially auto-calculate that based on the settings / capabilities in [Automate generation of the detail pages for Meltano plugins #253](https://github.com/meltano/hub/issues/253)
> * The URLs once you click into extractors have `/taps` and not `/extractors` (same for loaders / targets)

@tayloramurphy thanks!

I removed the optimized tag because I didnt think it was relevant anymore since all plugins can be added from Meltano now.

I'll look into the tap/extractor thing, I thought I fixed that at the end of the day yesterday.

PR #523, opened 2022-06-02
Hub classic navbar search updates

alexmarple:
This PR will add a dropdown for all plugins as well as a navbar level search feature for all plugins.


alexmarple:
@tayloramurphy oh yeah that's just me mucking around. if you grab latest that'll be gone. i'm still working on the logic of searching ALL collections. i might need to upgrade jekyll to get access to a helper.

On another note! if you have suggestions about what links should be in the nav, i'm all ears. should we have a maintainers page link in there? i think @pnadolny13 said he added one in one of his PRs?

tayloramurphy:
@alexmarple can we do:

*  Plugins - with the current dropdowns you have
* Maintainers - straight link
* Singer - dropdown to Taps, Targets, SDK, Spec, Docs
* Meltano - in the bubble format

Let's start with that as an iteration. We can change once we merge depending on how it feels, but if the pattern is there it's easy to change.

tayloramurphy:
@alexmarple I can't build this locally. Douwe made some updates to update jekyll and Ruby so things built faster and Netlify was happy in #543 Might need to merge in the latest from `develop`

alexmarple:
@tayloramurphy @pnadolny13 @aaronsteers @DouweM this is _technically_ good to go but i'm still trying to figure out the right style for that results list. if you have a suggestion, i'm all ears!

tayloramurphy:
@alexmarple I can leave a more detailed note in the morning but this is coming together nicely. Quick things I saw:

* The headers are completely gone on the singer pages. that's likely due to some logic about checking whether a page is singer related or not - I think we coudl probably remove that logic if the header stays consistent.
* the `Singer` link in the navbar should go to the resources page
* for the styling, no strong opinions yet, but it might be nice if the other plugin types completely disappeared if there are no results

I'd like others to chime in on the header options too

DouweM:
@alexmarple Thanks, I like it! I agree with Taylor that plugin categories should disappear when there are no matching results. Right now if you look for "great expectations", the actual utility is out of view and only the file bundle shows up. It would help to list Files last.

- [ ] Hide plugin categories without results
- [ ] List Files last

One more thing I noticed:
- [ ] The results list is only visible as long as your mouse cursor is over the search field or the list itself, meaning that if you search and want to move your mouse diagonally to the result and you accidentally cross the left border of the search field BEFORE crossing the top border of the result list, the list disappears before you can click your result :)

pnadolny13:
@alexmarple this looks cool! Other than the feedback already listed here are a few:

- [ ] Similar to Taylor's comment in https://github.com/meltano/hub/pull/523#issuecomment-1148107446 I think the `Plugins` navbar item should be clickable and link to the home page (i.e. `/`) with the plugin summary grid.
- [ ] Does the search use the keywords list in the plugin definition? I think this is more of a follow on but in case its simple, I would expect to search for `meltano_sdk` or `SDK` or `meltano sdk` to snag all with that keyword in their keywords list.

alexmarple:
@pnadolny13 i'm not currently using the keywords. just name, label, and namespace. i will add keywords as obviously those should be in the mix for searching.

DouweM:
@alexmarple I'm getting a syntax error

```
Uncaught SyntaxError: missing : after property id[main.js:109:10](http://localhost:4000/js/main.js)
```

_includes/header.html
```diff
@@ -11,6 +11,31 @@
         </svg>
       </a>
 
+      {% include navbar_search.html %}
+
+      <div class="dropdown">
+        <a class="page-link" href="{{site.url}}">Plugins</a>
+        <div class="dropdown-content">
+          <a href="{{ "/extractors/" | prepend: site.baseurl | prepend: site.url }}">Extractors</a>
```
tayloramurphy, line 19:
```suggestion
          <a href="{{ "/extractors/"  }}">Extractors</a>
```

all of these prepends can be dropped

tayloramurphy (approved):
Approving from a front-end experience perspective. We can make changes in a fast follow if needed.

Issue #522, opened 2022-06-02
Markdown page refactoring


aaronsteers:
Markdown page refactoring to spin out from this issue as separate work item and assignee:

- [x] Rename/move all Meltano plugin markdown files to `_data/meltano/<plugin_type>/<plugin_name>/<variant>-usage.md`.
- [x] Delete from the Meltano plugin `usage.md` files anything that will be autogenerated.
- [x] Move any sections that must come before installation to a new `_data/meltano/<plugin_type>/<plugin_name>/<variant>-prereqs.md`

_Originally posted by @aaronsteers in https://github.com/meltano/hub/issues/253#issuecomment-1145078283_

aaronsteers:
@pnadolny13 and @tayloramurphy - Talking this over with @alexmarple, we decided we needed to change the naming convention. Please see above for location and names.

"top" and "bottom" are confusing and not accurate - in the case of 'top' not actually being at the top and 'bottom' having no meaningful interpretation out of context. ("target-snowflake/meltano-usage.md" is readme-type info of some sort vs ""target-snowflake/meltano-bottom.md" you wouldn't know the content is actually a readme) unless you had some instructions saying so.

Also, when speaking colloquially: "you can put additional instructions in a bottom.md file" vs "you can put additional instructions in a usage.md file".

Related to https://github.com/meltano/hub/issues/253

aaronsteers:
Also - We should just set the expectations here, and in the contributor guide, that each markdown file should start with a L2 markdown header (`##`).

aaronsteers:
Agreed - and thanks, @tayloramurphy !

PR #525, opened 2022-06-02
Move singer-io/tap-adwords content

tayloramurphy:


tayloramurphy:
@pnadolny13 I started this branch to get the content changes for #522 

FYI @alexmarple you can use this as a starting point. We'll need your feedback on where this should live if the proposed format doesn't work cc @aaronsteers

tayloramurphy:
@DouweM @pnadolny13 this is ready for review

https://deploy-preview-525--meltano-hub.netlify.app/loaders/snowflake--datamill-co and https://deploy-preview-525--meltano-hub.netlify.app/targets/snowflake--datamill-co are good examples of how the `prereq` and `usage` keys render on Meltano plugins vs Singer connectors. 

I don't expect a full review of the content, I'm fairly happy with that, but a glance at it would be good for any obvious things.

_data/meltano/extractors/tap-bing-ads/singer-io.yml
```diff
@@ -144,30 +144,44 @@ settings_group_validation:
...
+  description: See https://docs.microsoft.com/en-us/advertising/guides/get-started#get-developer-token.
 - name: oauth_client_id
+  label: OAuth Client ID
   env_aliases:
   - OAUTH_BING_ADS_CLIENT_ID
-  label: OAuth Client ID
   kind: password
+  description: See https://docs.microsoft.com/en-us/advertising/guides/get-started#quick-start-production.
 - name: oauth_client_secret
+  label: OAuth Client Secret
```
pnadolny13, line 159:
```suggestion
```
Looks like the label is duplicated

schemas/extractors.schema.json
```diff
@@ -29,6 +29,8 @@
                 "maintenance_status": {},
                 "domain_url": {},
                 "keywords": {},
+                "usage": {},
```
pnadolny13, line 32: @tayloramurphy are these keys expected to also be returned by the API? If not we can add them to https://github.com/meltano/hub/blob/40ae36b69c691c065aacafe75b10252d22ef2f64/_plugins/meltano_json_generator.rb#L68 to exclude them from the generated static files. Also this makes me wonder if we should split the schemas between plugin and API in a follow on because theyre starting to diverge with some keys excluded. The schemas were orignally created to valid the API static files but overlapped with the plugin files also, but since we have a few keys were dropping it no longer validates perfectly i.e. validation would pass even though we dont allow this key in the API.

pnadolny13: Also if these are used for validating the plugin definition the maybe we should add them to the others as well because they might need custom usage notes too.

pnadolny13 (changes requested):
@tayloramurphy this looks great! I left a suggestion for a duplicate label and a potential update to exclude the new `usage`/`prereq` keys from the API json.

schemas/extractors.schema.json
tayloramurphy: @pnadolny13 let's leave them in for now since they are relevant instructions similar to the settings instructions.

I plan on adding them to the other plugins in #550

pnadolny13 (approved):
@tayloramurphy looks good to me!

PR #526, opened 2022-06-03
Limit filtering of plugins only to extractors/loaders/taps/targets

tayloramurphy:
This PR removes the search bar from every page except Extractors, Loaders, Taps, and Targets.


alexmarple (approved):
Checks out

Issue #219, opened 2022-03-24
Hub SSOT - Do we want to support package + metadata versioning?


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/hub/-/issues/223

Originally created by @pnadolny13 on 2022-03-24 14:17:25

---

First proposed in https://gitlab.com/meltano/hub/-/issues/180.

If we allow users to pin a version of their package, they need to be able to pin the metadata at that point in time too as metadata changes with the code. If a user has a 0.9.0 package version installed and a breaking change (say a new required setting was added) was introduced to the 1.0.0 (latest) version along with metadata updates, they cant just pull the latest metadata or their pipeline will break.

This may be resolved by the lock mechanism https://gitlab.com/meltano/meltano/-/issues/3031 but we should decide if we only want to the hub to represent the "latest" or "current" metadata.

This decision affects the way we implement the V2 API and the organization of the files in the hub.

MeltyBot:
[View 11 previous comments from the original issue on GitLab](https://gitlab.com/meltano/hub/-/issues/223)

tayloramurphy:
Notes from chat with AJ:

* supported Meltano versions: `meltano-version: ~2.5.0`
* Columns: date changed, meltano version constraints, comment of update type

Figure out how to automate on Hub, then figure out how Meltano can resolve it.

PR #528, opened 2022-06-03
Check if settings or capabilities are missing and add instructions cta

### LINKED ISSUE #527: Check if settings or capabilities are missing and add instructions / CTA
- [ ] Add logic to the script that puts a warning on the page if settings is missing, along with instructions and CTA for contributions.
- [ ] Add logic to the script that puts a warning on the page if capabilities is missing, along with instructions and CTA for contributions.

tayloramurphy:
Closes #527 

This PR:

* Removes the UI instructions from all connectors
* Updates the Settings / Capabilities to be size aware and display a CTA if too small
* Updates the YAML file link and sets it as a variable
Screen Shot 2022-06-03 at 10 24 29 AM
Screen Shot 2022-06-03 at 10 24 36 AM

### DIFF
--- a/_layouts/meltano_plugin.html
+++ b/_layouts/meltano_plugin.html
@@ -293,6 +293,7 @@ https://github.com/meltano/hub/tree/develop/_data/meltano/{{plugin_type}}s/{{plu
   <a href="#looking-for-help">learn how to get help</a>.
 </p>
 
+{% unless plugin_type contains ["extractor","loader","tap","target"] %}
 <h2 id="capabilities">Capabilities</h2>
     
 {% if plugin.capabilities.size == 0 %}
@@ -334,6 +335,8 @@ https://github.com/meltano/hub/tree/develop/_data/meltano/{{plugin_type}}s/{{plu
 </p>
 
 {% endif %}
+
+{% endunless %}
 
 {% if plugin.settings.size > 0 %}
 
--- a/_layouts/meltano_plugin.html
+++ b/_layouts/meltano_plugin.html
@@ -298,7 +298,7 @@ https://github.com/meltano/hub/tree/develop/_data/meltano/{{plugin_type}}s/{{plu
 {% if plugin.capabilities.size == 0 %}
 <p>
   <code>{{plugin.name}}</code> does not have any capabilities defined in its metadata.
-  Please considering adding them by making a PR to the
+  Please considering adding them by making a pull request to the
   <a href="{{yaml_file}}">YAML file</a>
   that defines the capabilities for this {{plugin_type}}.
 </p>
@@ -307,8 +307,8 @@ https://github.com/meltano/hub/tree/develop/_data/meltano/{{plugin_type}}s/{{plu
 
   {% if plugin.capabilities.size < 3 %}
     <p>
-      <code>{{plugin.name}}</code> may not have the full set of capabilities defined in its metadata.
-      Please considering adding them by making a PR to the
+      <code>{{plugin.name}}</code> may not have its full set of capabilities defined in its metadata.
+      Please considering adding them by making a pull request to the
       <a href="{{yaml_file}}">YAML file</a>
       that defines the capabilities for this {{plugin_type}}.
     </p>
@@ -328,7 +328,7 @@ https://github.com/meltano/hub/tree/develop/_data/meltano/{{plugin_type}}s/{{plu
 {% if plugin.settings.size == 0 %}
 <p>
   <code>{{plugin.name}}</code> does not have any settings defined in its metadata. 
-  Please considering adding them by making a PR to the 
+  Please considering adding them by making a pull request to the 
   <a href="{{yaml_file}}">YAML file</a>
   that defines the settings for this {{plugin_type}}.
 </p>
--- a/_layouts/meltano_plugin.html
+++ b/_layouts/meltano_plugin.html
@@ -298,20 +298,20 @@ https://github.com/meltano/hub/tree/develop/_data/meltano/{{plugin_type}}s/{{plu
     
 {% if plugin.capabilities.size == 0 %}
 <p>
-  <code>{{plugin.name}}</code> does not have any capabilities defined in its metadata.
+  <code>{{plugin.name}}</code> does not have any <a href="/singer/docs#singer-connector-capabilities">capabilities</a> defined in its metadata.
   Please considering adding them by making a pull request to the
   <a href="{{yaml_file}}">YAML file</a>
-  that defines the capabilities for this {{plugin_type}}.
+  that defines the <a href="/singer/docs#singer-connector-capabilities">capabilities</a> for this {{plugin_type}}.
 </p>
 
 {% else %} 
 
   {% if plugin.capabilities.size < 3 %}
     <p>
-      <code>{{plugin.name}}</code> may not have its full set of capabilities defined in its metadata.
+      <code>{{plugin.name}}</code> may not have its full set of <a href="/singer/docs#singer-connector-capabilities">capabilities</a> defined in its metadata.
       Please considering adding them by making a pull request to the
       <a href="{{yaml_file}}">YAML file</a>
-      that defines the capabilities for this {{plugin_type}}.
+      that defines the <a href="/singer/docs#singer-connector-capabilities">capabilities</a> for this {{plugin_type}}.
     </p>
   {% endif %}
 
--- a/_layouts/meltano_plugin.html
+++ b/_layouts/meltano_plugin.html
@@ -336,9 +336,7 @@ https://github.com/meltano/hub/tree/develop/_data/meltano/{{plugin_type}}s/{{plu
   that defines the settings for this {{plugin_type}}.
 </p>
 
-{% endif %}
-
-{% if plugin.settings.size > 0 %}
+{% else plugin.settings.size > 0 %}
 
 {% if plugin.settings_group_validation.size == 1 %}
 <p>
--- a/_layouts/meltano_plugin.html
+++ b/_layouts/meltano_plugin.html
@@ -299,18 +299,23 @@ https://github.com/meltano/hub/tree/develop/_data/meltano/{{plugin_type}}s/{{plu
 {% if plugin.capabilities.size == 0 %}
 <p>
   <code>{{plugin.name}}</code> does not have any <a href="/singer/docs#singer-connector-capabilities">capabilities</a> defined in its metadata.
-  Please considering adding them by making a pull request to the
+  Please consider adding them by making a pull request to the
   <a href="{{yaml_file}}">YAML file</a>
   that defines the <a href="/singer/docs#singer-connector-capabilities">capabilities</a> for this {{plugin_type}}.
 </p>
 
 {% else %} 
 
-  {% if plugin.capabilities.size < 3 %}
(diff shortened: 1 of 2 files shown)


_layouts/meltano_plugin.html
```diff
@@ -326,21 +295,47 @@ <h3 id="next-steps">Next steps</h3>
...
-{% if plugin.settings.size > 0 %}
 <h2 id="settings">Settings</h2>
+  
+{% if plugin.settings.size == 0 %}
+<p>
+  <code>{{plugin.name}}</code> does not have any settings defined in its metadata. 
+  Please considering adding them by making a PR to the 
+  <a href="{{yaml_file}}">YAML file</a>
+  that defines the settings for this {{plugin_type}}.
+</p>
```
aaronsteers, line 334: @tayloramurphy - Until https://github.com/meltano/meltano/issues/2717 ships, can we add instructions here to use `--store=dotenv` or `--store=db` for any sensitive values?

Otherwise we're going to immediately have high risk of user creds being compromised in git.

https://github.com/meltano/meltano/issues/2717#issuecomment-1146092980

aaronsteers (changes requested):
Just one change requested regarding warning for sensitive config values when no settings are defined.

_layouts/meltano_plugin.html
tayloramurphy: @aaronsteers there are no instructions on the Hub for how to configure with Meltano. We link users to https://docs.meltano.com/guide/configuration when there _are_ settings. I'm not sure adding instructions on `meltano config` specifically here makes sense.

_layouts/meltano_plugin.html
```diff
@@ -326,21 +295,47 @@ <h3 id="next-steps">Next steps</h3>
...
-{% if variant.capabilities.size == 0 %}
+{% if plugin.capabilities.size == 0 %}
 <p>
-  Check the
-  <a href="{{variant.repo}}#readme" target="_blank">README and code in the repository</a>
-  for more information on capabilities for this {{plugin_type}}.
+  <code>{{plugin.name}}</code> does not have any capabilities defined in its metadata.
+  Please considering adding them by making a PR to the
+  <a href="{{yaml_file}}">YAML file</a>
+  that defines the capabilities for this {{plugin_type}}.
```
DouweM, line 303: Can we (now or soon after 2.0) link to a doc describing what capabilities are, and how to find out which ones a tap/target supports? The closest thing we have today is https://docs.meltano.com/guide/integration#validate-tap-capabilities and https://docs.meltano.com/contribute/plugins#how-to-test-a-tap, but neither are as clear as they could be.

Edit: I just found https://hub.meltano.com/singer/docs#singer-connector-capabilities! That's not super clear either though, since "Log Based" for example isn't a capability keyword. And `discovery` should be `discover`, shouldn't it?

_layouts/meltano_plugin.html
```diff
@@ -326,21 +295,47 @@ <h3 id="next-steps">Next steps</h3>
 
 <h2 id="capabilities">Capabilities</h2>
     
-{% if variant.capabilities.size == 0 %}
+{% if plugin.capabilities.size == 0 %}
```
DouweM, line 298: Can we only show this for taps? SQLFluff and other utilities without capabilities are fine.

_layouts/meltano_plugin.html
```diff
@@ -326,21 +295,47 @@ <h3 id="next-steps">Next steps</h3>
...
+  <code>{{plugin.name}}</code> does not have any capabilities defined in its metadata.
+  Please considering adding them by making a PR to the
+  <a href="{{yaml_file}}">YAML file</a>
+  that defines the capabilities for this {{plugin_type}}.
 </p>
-{% else %} <ul> {% for capability in plugin.capabilities %} 
+
+{% else %} 
+
+  {% if plugin.capabilities.size < 3 %}
```
DouweM, line 308: For lots of taps, 2 capabilities are perfectly fine, and I don't think we should suggest to users that it's not. 

Since this is to catch taps that now have `state` and `catalog` as defaults even if we haven't verified that, I think we're better off removing the defaults or making it explicit with a property like `metadata_verified` (that could be `false` by default and set to `true` on the subset that are). We can then also clarify that this tap may not work out-of-the-box, and what to do if they hit an issue. It may not just be missing capabilities, but could also have incorrect ones (like `catalog` instead of `properties`).

I think it's important that people know from the plugin page whether they can expect the tap to work out of the box, or not. These paragraphs hint at that, but I don't think they set expectations sufficiently yet, or help a user address potential issues with taps with missing/broken metadata.

_layouts/meltano_plugin.html
```diff
@@ -326,21 +295,47 @@ <h3 id="next-steps">Next steps</h3>
...
+    <p>
+      <code>{{plugin.name}}</code> may not have the full set of capabilities defined in its metadata.
```
DouweM, line 310:
```suggestion
      <code>{{plugin.name}}</code> may not have its full set of capabilities defined in its metadata.
```

_layouts/meltano_plugin.html
```diff
@@ -326,21 +295,47 @@ <h3 id="next-steps">Next steps</h3>
...
 <h2 id="capabilities">Capabilities</h2>
     
-{% if variant.capabilities.size == 0 %}
+{% if plugin.capabilities.size == 0 %}
 <p>
-  Check the
-  <a href="{{variant.repo}}#readme" target="_blank">README and code in the repository</a>
-  for more information on capabilities for this {{plugin_type}}.
+  <code>{{plugin.name}}</code> does not have any capabilities defined in its metadata.
+  Please considering adding them by making a PR to the
```
DouweM, line 301:
```suggestion
  Please considering adding them by making a pull request to the
```

_layouts/meltano_plugin.html
```diff
@@ -326,21 +295,47 @@ <h3 id="next-steps">Next steps</h3>
...
+
+  {% endif %}
 
-{% if plugin.settings.size > 0 %}
 <h2 id="settings">Settings</h2>
+  
+{% if plugin.settings.size == 0 %}
+<p>
+  <code>{{plugin.name}}</code> does not have any settings defined in its metadata. 
+  Please considering adding them by making a PR to the
```
DouweM, line 331:
```suggestion
  Please considering adding them by making a pull request to the 
```

_layouts/meltano_plugin.html
```diff
@@ -326,21 +295,47 @@ <h3 id="next-steps">Next steps</h3>
...
+<p>
+  <code>{{plugin.name}}</code> does not have any settings defined in its metadata. 
+  Please considering adding them by making a PR to the 
+  <a href="{{yaml_file}}">YAML file</a>
+  that defines the settings for this {{plugin_type}}.
+</p>
+
+{% endif %}
+
+{% if plugin.settings.size > 0 %}
```
DouweM, line 338: This can become an `else`

DouweM (changes requested):

DouweM:
@tayloramurphy Some of what I suggested can be done later (like better docs to help people address missing capabilities/settings), but I think it's important that people don't lose confidence in the Hub because we're vague about whether plugins will work well or not.

_layouts/meltano_plugin.html
tayloramurphy: Resolved via 772fa1ba19c2ce878d480ce03c63a628e39a3fa5

_layouts/meltano_plugin.html
tayloramurphy: Resolved via 772fa1ba19c2ce878d480ce03c63a628e39a3fa5

_layouts/meltano_plugin.html
```diff
@@ -326,21 +295,47 @@ <h3 id="next-steps">Next steps</h3>
...
 
-{% if plugin.settings.size > 0 %}
 <h2 id="settings">Settings</h2>
+  
+{% if plugin.settings.size == 0 %}
+<p>
+  <code>{{plugin.name}}</code> does not have any settings defined in its metadata. 
+  Please considering adding them by making a PR to the 
+  <a href="{{yaml_file}}">YAML file</a>
+  that defines the settings for this {{plugin_type}}.
```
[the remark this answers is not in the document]
tayloramurphy: We don't. I made https://github.com/meltano/hub/issues/534 to track

_layouts/meltano_plugin.html
tayloramurphy: @DouweM I understand the concerns, I just don't know how we're going to keep that up to date over the long run. A static `metadata_verified` will work for the short-term but is likely to be wrong as updates to the underlying package are created. We have https://github.com/meltano/hub/issues/219 which would give us a better indication of metadata versioning on a more automated basis. 

The original capabilities that I set when we first launched the Hub were `catalog` and `discover`. What do you think about softer language that only triggers when those 2 capabilities are set AND there are no settings?

tayloramurphy: @DouweM take a look at the latest changes f78fe9048830759c804262e2fff6546a98fb752c

_layouts/meltano_plugin.html
DouweM: @tayloramurphy The current Markdown pages have `meltano config` instructions: https://hub.meltano.com/extractors/gitlab#how-to-use-1. Is auto-generating those sections not part of https://github.com/meltano/hub/issues/253? We don't want to lose valuable guidance on configuration.

tayloramurphy: @DouweM we are showing the config instructions when settings are present as you would expect. The argument I was making here is that when no settings are present showing config instructions doesn't feel right.

aaronsteers (approved):
@tayloramurphy - I'm not sure I'm comfortable with this being the final version, but this is clearly better than prior so it makes for a stable iteration.

I don't know if a user looking at this would be able to take the correct actions to properly configure their plugins, and I'll try to come back to this next week if time allows.

_layouts/meltano_plugin.html
DouweM: @tayloramurphy I like https://github.com/meltano/hub/issues/219 and this workaround, thanks!

singer/docs.md
```diff
@@ -73,7 +73,7 @@ For more details see the <a href="{{ "/singer/spec#catalog-files" | prepend: sit
 The tap accepts the legacy `--properties` argument referencing a file that defines the structure of one or many data streams.
 For more details see the <a href="{{ "/singer/spec#catalog-files" | prepend: site.baseurl | prepend: site.url }}">Catalog Files section</a> in our Singer Spec interpretation.
 
-#### Discovery
+#### Discover
 
 The tap accepts a `--discovery` argument that is used to generate a catalog files.
```
DouweM, line 78:
```suggestion
The tap accepts a `--discover` argument that is used to generate a catalog files.
```

DouweM (changes requested):

DouweM (approved):

DouweM:
@tayloramurphy Agreed with @aaronsteers, much better than what we had and we can always iterate from here. I think we can merge without waiting for Pat and Alex to approve as well. Looks like there's still a conflict to be resolved tthough.

alexmarple:
@tayloramurphy i haven't had a chance to look this over thoroughly but i trust that aj and douwe know what they're talking about!

PR #530, opened 2022-06-03
Add superset utility

pnadolny13:
This is a simple first iteration.

- It allows you to set driver dependencies via settings to allow environment level configs.
- load_examples will populate your instance with their default example datasets, dashboards, and charts
- `config_path` points to the superset_config.py added from the file bundle in https://github.com/meltano/files-superset/pull/12. This lets users set whatever custom configs they want. Superset doesnt work like Airflow in the sense that all configs can be accepted as env vars. Having this config file is a simple way to give flexibility for the first iteration.
- The reason for adding the markupsafe to the pip_url is because without it I get `ImportError: cannot import name 'soft_unicode' from 'markupsafe'` and an issue is already being tracked in the Superset repo https://github.com/apache/superset/issues/19150. I'm open to other solutions though.

If theres other content we want to add to the plugin page I can write the markdown once https://github.com/meltano/hub/issues/522 is done. Let me know if you have particular idea for what to include.

```bash
meltano add utility superset
# Optional settings to either get postgres driver for connecting to local postgres or loading examples
# meltano config superset set additional_dependencies "psycopg2==2.9.3"
# meltano config superset set load_examples true
meltano invoke superset:ui
```


_data/meltano/utilities/superset/apache.yml
```diff
@@ -0,0 +1,32 @@
+name: superset
+namespace: superset
+label: Apache Superset
+description: A modern, enterprise-ready business intelligence web application.
+docs: https://docs.meltano.com/guide/orchestration
+repo: https://github.com/apache/superset
+variant: apache
+pip_url: apache-superset==1.5.0 markupsafe==2.0.1
```
DouweM, line 8:
```suggestion
pip_url: apache-superset==1.5.0 markupsafe==2.0.1 # `markupsafe` is listed explicitly to resolve https://github.com/apache/superset/issues/19150 
```

_data/meltano/utilities/superset/apache.yml
```diff
@@ -0,0 +1,32 @@
...
+requires:
+  files:
+  - name: files-superset
+    variant: meltano
+settings:
+- name: home
+  value: $MELTANO_PROJECT_ROOT/.meltano/run/superset
```
DouweM, line 15: @pnadolny13 Where is this used? I don't see it referenced in https://github.com/meltano/files-superset/pull/12. Is this to set the `SUPERSET_HOME` env var that it reads from?

Note that in the case of Airflow, `AIRFLOW_HOME` is automatically set at runtime, not through a setting, since it shouldn't be overridden: https://github.com/meltano/meltano/blob/main/src/meltano/core/plugin/airflow.py#L19

_data/meltano/utilities/superset/apache.yml
```diff
@@ -0,0 +1,32 @@
...
+  description: The home directory where Superset will store data.
+- name: additional_dependencies
```
DouweM, line 17: I'd rather remove this and instruct people to add dependencies to `pip_url`, like we've done with Airflow: https://docs.meltano.com/guide/production#airflow-orchestrator:

> For this to work, the [psycopg2 package](https://pypi.org/project/psycopg2/) will also need to be installed alongside [apache-airflow](https://pypi.org/project/apache-airflow/), which you can realize by adding psycopg2 to airflow’s pip_url in your [meltano.yml project file](https://docs.meltano.com/concepts/project#meltano-yml-project-file) (e.g. pip_url: psycopg2 apache-airflow) and running [meltano install orchestrator airflow](https://docs.meltano.com/reference/command-line-interface#install).

Otherwise these dependencies will be reinstalled every time Superset is initialized.

_data/meltano/utilities/superset/apache.yml
```diff
@@ -0,0 +1,32 @@
...
+  description: An space separated list of python dependencies to include as part of startup.
+    A list of database driver dependencies can be found here https://superset.apache.org/docs/databases/installing-database-drivers
+- name: load_examples
+  kind: boolean
+  value: false
+  description: A boolean whether to load Superset's example datasets and dashboards. Warning, this takes a few mins to load.
```
DouweM, line 23: I don't think it makes sense for this to be a setting and to run the command every time on initialization, as it's something a user would only want to do once. We're better off instructing them to run `meltano invoke superset load_examples` manually if they want this.

_data/meltano/utilities/superset/apache.yml
```diff
@@ -0,0 +1,32 @@
...
+- name: config_path
+  value: $MELTANO_PROJECT_ROOT/analyze/superset/superset_config.py
+  description: The path to the superset_config.py file used to override config settings.
+    See https://superset.apache.org/docs/installation/configuring-superset for more details on configuration options.
+commands:
+  ui:
+    args: analyze/superset/superset-init.sh
+    executable: /bin/bash
+logo_url: /assets/logos/utilities/superset.png
```
DouweM, line 32: I don't think we need an explicit logo URL if it follows the expected pattenr.

_data/meltano/utilities/superset/apache.yml
```diff
@@ -0,0 +1,32 @@
...
+  value: false
+  description: A boolean whether to load Superset's example datasets and dashboards. Warning, this takes a few mins to load.
+- name: config_path
+  value: $MELTANO_PROJECT_ROOT/analyze/superset/superset_config.py
+  description: The path to the superset_config.py file used to override config settings.
+    See https://superset.apache.org/docs/installation/configuring-superset for more details on configuration options.
+commands:
+  ui:
+    args: analyze/superset/superset-init.sh
+    executable: /bin/bash
```
DouweM, line 31: What if a machine doesn't have bash? :)

Could we set `analyze/superset/superset-init.sh` as the `executable`?

_data/meltano/utilities/superset/apache.yml
```diff
@@ -0,0 +1,32 @@
...
+  description: The home directory where Superset will store data.
+- name: additional_dependencies
+  description: An space separated list of python dependencies to include as part of startup.
+    A list of database driver dependencies can be found here https://superset.apache.org/docs/databases/installing-database-drivers
+- name: load_examples
+  kind: boolean
+  value: false
+  description: A boolean whether to load Superset's example datasets and dashboards. Warning, this takes a few mins to load.
+- name: config_path
+  value: $MELTANO_PROJECT_ROOT/analyze/superset/superset_config.py
```
DouweM, line 25: dbt lives in `/transform`, and Airflow DAGs live in `/orchestrate`, without `/dbt` or `/airflow` added into the path. What do you think about being consistent and using `/analyze` here (and above, and in the file bundle) without `/superset`? 

Or would it be better to change the others to use the plugin name explicitly in the path as well, in situations where you'd have multiple plugins of a type?

@tayloramurphy Wdyt?

DouweM (changes requested):

_data/meltano/files/superset/meltano.yml
```diff
@@ -0,0 +1,8 @@
+name: superset
+namespace: superset
+label: Apache Superset
+description: A modern, enterprise-ready business intelligence web application.
+repo: https://github.com/meltano/files-superset
+pip_url: git+https://github.com/meltano/files-superset.git
+variant: meltano
+logo_url: /assets/logos/files/superset.png
```
DouweM, line 8: Not sure if this is needed if the name matches the expected pattern.

_data/meltano/utilities/superset/apache.yml
```diff
@@ -0,0 +1,32 @@
+name: superset
```
DouweM, line 1: We shouldn't merge this until we have a have first-user instructions around creating an admin user, overriding `secret_key` etc.

DouweM (changes requested):

_data/meltano/files/superset/meltano.yml
aaronsteers: Regarding logo_url, I don't know how the api generation script leverages this, but the goal is to have an fully qualified URL exposed in the API endpoint. Towards this end, I do prefer having the logo_url everywhere just so we're not relying on over-relying on convention.

If the API generation process is able to adapt and map correctly, I'd be okay with any of these three:

```
logo_url: superset.png                     # when no path part is provided, find the file in `/assets/logos/files/`
logo_url: /assets/logos/files/superset.png # abs url will be https://raw.githubusercontent.com/meltano/hub/main/assets/logos/files/airflow.png
logo_url: https://raw.githubusercontent.com/meltano/hub/main/assets/logos/files/airflow.png
```

Sidebar (and I'll open an issue for this): I don't think image storage should be brand-specific or logo-specific versus plugin_type-specific. I think we should consider `/assets/logos/` instead of `/assets/logos/files/`.

For instance, `tap-snowflake`, `target-snowflake`, `dbt-snowflake`, and `files-dbt-snowflake` might all use `https://raw.githubusercontent.com/meltano/hub/main/assets/logos/snowflake.png`. If we establish that logos are (by default) in `/assets/logos/`, then all we need here is `snowflake.png`. And if/when we need to, we can use slightly varying logo files like `dbt-snowflake.png` or `snowflake.jpg`, and this gives us control, flexibility, and berevity.

aaronsteers: https://github.com/meltano/hub/issues/545#issue-1260639744

_data/meltano/utilities/superset/apache.yml
pnadolny13: @DouweM yes its tells Superset where to write the system db and any other files. Its not explicitly referenced in the start up script in https://github.com/meltano/files-superset/pull/12 but Superset itself uses it. If I understand correctly, you're saying I should hard code it like `SUPERSET_HOME=$MELTANO_PROJECT_ROOT/.meltano/run/superset` in the startup script rather than allowing users to override in settings? Is that right?

_data/meltano/utilities/superset/apache.yml
pnadolny13: @DouweM yeah I noticed that too, I think its better to namespace them. Probably for other cases too but especially for someone exploring plugins locally they could have multiple analyze plugins installed, maybe lightdash and superset to compare, and they'd get this shared directory with all files lumped together. Same with dbt, like they could have multiple warehouses and include dbt-redshift and dbt-snowflake.

_data/meltano/files/superset/meltano.yml
pnadolny13: @aaronsteers yeah the API generation script does leverage it and adds the base url to it like `https://hub.meltano.com` or `http://localhost:4000` or similar, depending on where you're running it.

> Sidebar (and I'll open an issue for this): I don't think image storage should be brand-specific or logo-specific versus plugin_type-specific. I think we should consider /assets/logos/ instead of /assets/logos/files/.

Yeah I also agree we could definitely share the logos better, we can do that in a follow on though.

_data/meltano/utilities/superset/apache.yml
pnadolny13: @DouweM I wasn't able to get it to work without `/bin/bash`. I'm not super aware of the full ecosystem but dont most Linux machines have bash? I know this wouldnt work for windows though. What do you think is best?

I was actually wondering if it would be better to use python for these scripts. All users need Python and our ecosystem  (and I) are way more comfortable with python anyways 😄 .

_data/meltano/utilities/superset/apache.yml
tayloramurphy: @DouweM I'm with @pnadolny13 on this one. Namespacing the directories will likely be preferred going forward. Especially if we want people to maintain everything in one repo, it's not a good assumption that they're going to have only a single tool for each top-level directory.

DouweM:
To document:
- [ ] Adding dependencies to `pip_url`: https://superset.apache.org/docs/installation/installing-superset-from-scratch/#os-dependencies
- [ ] meltano config superset set secret_key $(openssl rand -base64 42)
- [ ] meltano invoke superset:create-admin
- [ ] meltano invoke superset:load_examples

To do:
- [ ] Superset detail page says it's a "Meltano loader". May already be fixed in https://github.com/meltano/hub/pull/541.

tayloramurphy:
> Superset detail page says it's a "Meltano loader". May already be fixed in https://github.com/meltano/hub/pull/541.

@DouweM I'm realizing that's not going to be the case. It seems like everything that's not an Extractor or Loader is still built from a markdown file. I can take a stab at seeing if I can update https://github.com/meltano/hub/pull/541/files to be generic for all plugin types. Let me see what breaks when I remove those files and see what needs to be updated. I can update in #541

tayloramurphy (approved):
Great, stable iteration. Approved ✅

DouweM (approved):

PR #538, opened 2022-06-03
add file bundle requires attributes

pnadolny13:


aaronsteers:
@pnadolny13 - While name doesn't have to be unique across plugin types, I think it's helpful for it to be so. This avoids collisions and having to inform users how to resolve them by providing an additional `--plugin-type` argument.

Can we keep with the name being `files-<something>` such as `files-airflow` and similar?

pnadolny13:
> Can we keep with the name being `files-<something>` such as `files-airflow` and similar?

@aaronsteers yeah I'm good with that. I just took them as they were named [in discovery.yml](https://github.com/meltano/meltano/blob/eee011d2f7e1048d1326cea5cf348969cc591332/src/meltano/core/bundle/discovery.yml#L3493).

I also see that transforms do the same thing like for https://github.com/meltano/hub/blob/develop/_data/meltano/transforms/tap-adwords/meltano.yml the `name` is `tap-adwords` but the package is really `dbt-tap-adwords`, same with the pipelinewise variants like https://github.com/meltano/hub/blob/develop/_data/meltano/extractors/tap-postgres/transferwise.yml with `name` as `tap-postgres` and package `pipelinewise-tap-postgres`. I think pipelinewise might be an exception because they dont need to be independent plugins but still wanted to call it out.

aaronsteers:
@pnadolny13 - Yeah, my hunch is that this was to facilitate auto-adding based on same-name or same-namespace. Going forward, I think the plugin having a specific name is preferrable.

cc @tayloramurphy

aaronsteers (changes requested):
As discussed, requesting `files-*` names here. Thanks.

tayloramurphy:
@pnadolny13 adding my concurrence with @aaronsteers on the naming. `files-x` 👍

pnadolny13:
@aaronsteers @tayloramurphy this is updated to have the `files-` prefix as part of the name and I'd also call out that I kept the convention for namespaces to include underscores instead of dashes. So for example what was previously `namespace: great_expectations` became `namespace: files_great_expectations`.

tayloramurphy:
Fixed the merge conflicts. Merging with my superpowers so I can continue on #550

Issue #252, opened 2022-05-27
Update Airflow metadata to include airflow:ui command


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/hub/-/issues/257

Originally created by @tayloramurphy on 2022-05-27 13:39:24

---

From https://meltano.slack.com/archives/C02MBDUHCCD/p1653575513089299 we should have a simple `ui` command for Airflow

MeltyBot:
[View 2 previous comments from the original issue on GitLab](https://gitlab.com/meltano/hub/-/issues/257)

aaronsteers:
@pnadolny13 - Linking and closing, per #540. 🙌

PR #540, opened 2022-06-03
add airflow ui command

pnadolny13:
Closes https://github.com/meltano/hub/issues/252

The current implementation adds a `ui` command that just aliases webserver:

```
meltano invoke airflow:ui
# Is equivalent to the current
meltano invoke airflow webserver
```

Do we want more than this? It assumes that the user has already turned on the scheduler in the background. We could alternatively add a startup script similar to https://github.com/meltano/files-superset/blob/5a3c212933a2b76e09422ee3ab541afdf4fc6b98/bundle/analyze/superset/superset-init.sh if we wanted to start the scheduler in the background then start the webserver in the foreground but then we'd have have a way of shutting the scheduler down on exit. It wouldnt be all that hard but I wasnt sure what we were looking for.


DouweM:
@pnadolny13 Looks good! In production you wouldn't want your scheduler and webserver to run in the same container, so requiring them to be started separately is a good thing.

DouweM (approved):

tayloramurphy (approved):

PR #541, opened 2022-06-03
Update Singer-specific pages to be Meltano-neutral

tayloramurphy:
This PR updates the Singer pages on the develop branch to be more like what currently exists.

TODO:

- [ ] Update alternate variant URLs

### DIFF
--- a/_includes/setting.html
+++ b/_includes/setting.html
@@ -7,6 +7,8 @@
   {% endif %}
 </h3>
 
+{% if is_singer == false %}
+
 <ul>
   {% if plugin_name %} {% capture env_var %}{{plugin_name | replace: "-", "_" |
   upcase}}_{{setting.name | upcase}}{% endcapture %}
@@ -35,8 +37,7 @@
   <li>Default: <code>{{setting.value}}</code></li>
   {% endif %}
 </ul>
-{{ setting.description | markdownify }} {% if setting.documentation
-%}
+{{ setting.description | markdownify }} {% if setting.documentation%}
 <p>
   See
   <a href="{{setting.documentation}}" target="_blank"
@@ -47,12 +48,12 @@
 <h4 id="how-to-use-{{ setting.name | slugify }}">How to use</h4>
 
 <p>
-  Manage this setting using <a href="#using-meltano-ui">Meltano UI</a>,
+  Manage this setting using 
   <a
     href="https://docs.meltano.com/command-line-interface.html#config"
     taregt="_blank"
     ><code>meltano config</code></a
-  >, or an
+  > or an
   <a
     href="https://docs.meltano.com/configuration.html#configuring-settings"
     target="_blank"
@@ -70,4 +71,15 @@ endcase %}{% endcapture %}
 <pre><code>meltano config {{plugin_name}} set {{setting.name}} {{example_value}}
 
 export {{env_var}}={{example_value}}</code></pre>
+{% endif %}
+
+{% else %}
+
+<ul>
+  {% if setting.value != nil %}
+  <li>Default: <code>{{setting.value}}</code></li>
+  {% endif %}
+</ul>
+{{ setting.description | markdownify }}
+
 {% endif %}
--- a/_layouts/meltano_plugin.html
+++ b/_layouts/meltano_plugin.html
@@ -8,6 +8,35 @@ layout: plugin_page
 https://github.com/meltano/hub/tree/develop/_data/meltano/{{plugin_type}}s/{{plugin.name}}/{{plugin.variant}}.yml
 {% endcapture %}
 {% if plugin.metrics %} {% assign metrics = plugin.metrics %}{% endif %}
+{% assign is_singer = false %}
+{% if plugin_type == "tap" or plugin_type == "target" %} {% assign is_singer =  true %} {% endif %}
+
+{% if is_singer == true %}
+
+<p>
+  {% if plugin_type == "tap" %} The <code>{{plugin.name}}</code>
+  <a href="/singer/taps">Singer tap</a>
+  pulls data from
+  {% if plugin.entity_url %}
+  <a href="{{plugin.entity_url}}" target="_blank">{{plugin.label}}</a>
+  {% else %}
+  the {{ plugin.label }} {{ plugin.entity_type }}
+  {% endif %}
+  that can then be sent to a destination using a
+  <a href="/singer/targets">Singer target</a>. {% else %} The
+  <code>{{plugin.name}}</code>
+  <a href="/singer/targets">Singer target</a>
+  sends data into
+  {% if plugin.entity_url %}
+  <a href="{{plugin.entity_url}}" target="_blank">{{plugin.label}}</a>
+  {% else %}
+  the {{ plugin.label }} {{ plugin.entity_type }}
+  {% endif %}
+  after it was pulled from a source using a
+  <a href="/singer/taps">Singer tap</a>. {% endif %}
+</p>
+
+{% else %}
 
 <p>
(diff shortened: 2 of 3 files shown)


tayloramurphy:
@DouweM I could use a review on this. Locally I'm able to build it and the default variants work just like you'd expect. On this build though I'm seeing the wrong variant as the default page even though it's listed in the variants list as `(default)`. https://deploy-preview-541--meltano-hub.netlify.app/taps/auth0

Maybe it's something with my Ruby version locally? I'm using 2.7.

Otherwise I'm pretty happy with this PR - it's actually better than what's currently on the Hub! https://hub.meltano.com/taps/auth0

tayloramurphy:
I've updated Ruby to 3.1.2 and I see the same behavior currently on develop. Looks like it's perhaps a sorting thing?

tayloramurphy:
Found it! 🎉  6f0045e31ef928862e8af4b877ee5d1687a94f0b

tayloramurphy:
@DouweM @alexmarple this is ready for review. 

A nice side by side comparison to see current state vs what's in this PR are the Auth0 pages listed in https://github.com/meltano/hub/pull/541#issuecomment-1146514081

We're still building both sets of pages from a single HTML file which is fine I think for the launch and will change presumably with the Gridsome Redesign.

DouweM:
@tayloramurphy Comparing them side by side is pretty handy, I spotted a few differences already that I think we should fix:
- [ ] Page title says "Meltano tap" instead of "Singer tap"
- [ ] Navbar says "MeltanoHub" instead of "MeltanoHub for Singer"
- [ ] Description says "the Auth0" without a link instead of "Auth0" with a link
- [ ] Repository list lacks "Last Updated At"
- [ ] "Meltano Stats" are missing
- [ ] "Edit this page on GitLab" is missing

Except for the first 2, all of these changes would be good on the Meltano-specific pages as well. I'm actually not sure why these details went missing in the first place, if they're currently live on hub.meltano.com?

I also suggest changing the "Usage with Meltano" section to make the point of Meltano a little more clear:

> [Meltano](https://meltano.com) helps you manage your [configuration](https://docs.meltano.com/guide/configuration), [incremental replication](https://docs.meltano.com/guide/integration#incremental-replication-state), and [scheduled pipelines](https://docs.meltano.com/guide/orchestration#create-a-schedule). 
> 
> View the [Meltano-specific `tap-auth0` instructions](/extractors/auth0#getting-started) to learn more.

Note that the second link includes `#getting-started` so that the user is directed to the new section, and not a page that at first glance looks exactly like the page they just saw.

tayloramurphy:
> Page title says "Meltano tap" instead of "Singer tap"
> Navbar says "MeltanoHub" instead of "MeltanoHub for Singer"

I'm going to pop a separate issue to track that. That happened b/c of the consolidation of the Meltano and Singer pages into one so I'll need to spend more time on fixing that, but I'd like to keep it separate from the pure content part if that's ok. https://github.com/meltano/hub/issues/548

> Description says "the Auth0" without a link instead of "Auth0" with a link

Fixed

> Repository list lacks "Last Updated At"
> "Meltano Stats" are missing

This is because the local (and presumably Netlify) versions of the Hub are using a smaller static file to pull in metrics. The full deploy process gets everything from S3 (cc @pnadolny13 if I've mispoken here). If you look at https://deploy-preview-541--meltano-hub.netlify.app/taps/slack you'll see the metrics. 

> "Edit this page on GitLab" is missing

Fixed and changed to GitHub!

> I also suggest changing the "Usage with Meltano" section

Done!

_includes/setting.html
```diff
@@ -7,6 +7,8 @@ <h3 id="{{ setting.name | slugify }}">
   {% endif %}
 </h3>
 
+{% if is_singer == false %}
```
DouweM, line 10: Is `is_singer` actually available here? As far as I know, variables have to be passed into included partials explicitly and then referenced as `include.<var>` as in line 1 of this file.

_layouts/meltano_plugin.html
```diff
@@ -9,6 +9,44 @@
 {% endcapture %}
 {% if plugin.metrics %} {% assign metrics = plugin.metrics %}{% endif %}
 
+<!-- To determine if the current plugin is pure Singer or not -->
+{% assign is_singer = false %}
```
DouweM, line 13: We may be better off moving this logic to https://github.com/meltano/hub/blob/main/_plugins/meltano_plugin_page_generator.rb so that we can also use it to toggle the page title and header (https://github.com/meltano/hub/issues/548).

If we add it to `@data` there, it becomes available as `page.is_singer` here, and likely in `_includes/setting.html` as well -- although we may still need to pass it there explicitly.

DouweM (changes requested):
One last potential issue and a workaround that may also help with #548.

I also just noticed that the page descriptions on https://deploy-preview-541--meltano-hub.netlify.app/taps/auth0 is wrong:
> The open source Meltano loader sends data into Auth0 after it was pulled from a source using an extractor.

https://deploy-preview-541--meltano-hub.netlify.app/extractors/auth0 is right:
> The open source Meltano extractor pulls data from Auth0 that can then be sent to a destination using a loader.

We may only have `if extractor; else` logic, meaning that anything not an extractor gets the loader text.

_includes/setting.html
tayloramurphy: If you compare the settings on https://deploy-preview-541--meltano-hub.netlify.app/extractors/auth0 and https://deploy-preview-541--meltano-hub.netlify.app/taps/auth0 it does seem to work. 🤷

_layouts/meltano_plugin.html
tayloramurphy: @DouweM I was able to move the logic and update everything. Much easier to do it in the Ruby script!

I'm going to open another PR targeting this branch that will work to fix the other plugin pages. I've got a few commits ready but will have to wait until tomorrow to continue it.

pnadolny13 (approved):
@tayloramurphy thanks for the quick fixes! Looks good to me.

DouweM (changes requested):
- [ ] The page description incorrectly describes anything other than extractors as loaders

_layouts/meltano_plugin.html
```diff
@@ -5,41 +5,32 @@
 {% assign plugin_type = page.plugin_type %} {% assign plugin = page.plugin %} 
 {% assign repo_package_name = plugin.repo | split: "/" | last %}
 {% capture yaml_file %}
-https://github.com/meltano/hub/tree/develop/_data/meltano/{{plugin_type}}s/{{plugin.name}}/{{plugin.variant}}.yml
+https://github.com/meltano/hub/tree/develop/_data/meltano/{{page.meltano_type}}/{{plugin.name}}/{{plugin.variant}}.yml
 {% endcapture %}
 {% if plugin.metrics %} {% assign metrics = plugin.metrics %}{% endif %}
 
 <!-- To determine if the current plugin is pure Singer or not -->
```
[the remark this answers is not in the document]
tayloramurphy: Fixed

_layouts/default.html
```diff
@@ -3,8 +3,13 @@
   {% include head.html %}
 
   <body>
-    {% if page.header %} {% include header_{{page.header}}.html %} {% else %} {%
-    include header.html %} {% endif %}
+    {% if page.header %} 
+      {% include header_{{page.header}}.html %} 
+    {% elsif page.is_singer == true%}
```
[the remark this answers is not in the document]
tayloramurphy: Per our DMs, going to leave as is and merge.

tayloramurphy:
> The page description incorrectly describes anything other than extractors as loaders

Addressing this in #550

PR #544, opened 2022-06-04
ci: add back CI workflows

aaronsteers:


DouweM:
@aaronsteers Should go into `develop` instead of `main`, right?

.github/workflows/jekyll_test.yml
```diff
@@ -0,0 +1,64 @@
+name: Jekkyll test
```
DouweM, line 1:
```suggestion
name: Jekyll test
```

.github/workflows/jekyll_test.yml
```diff
@@ -0,0 +1,64 @@
...
+on:
+  pull_request: {}
+  push:
+    branches: [main]
+  workflow_dispatch:
+    inputs: {}
+
+jobs:
+  tests:
+    name: Jekkyll build test
```
DouweM, line 12:
```suggestion
    name: Jekyll build test
```

DouweM (approved):

aaronsteers:
> @aaronsteers Should go into `develop` instead of `main`, right?

Yes, it will need to get there. However, since I can't tell the status of that branch, I would like to see this working first.

aaronsteers:
@DouweM - Thanks for your review. 

Quick summary:

- This fails in `main` at the point of schema validation, due to `requires` being added ahead of its declaration in json schema. But it gets through the jekyll part with no problem: https://github.com/meltano/hub/runs/6734716316?check_suite_focus=true
- This succeeds in `develop` branch also.
- Runtime on both of these is <3 minutes.

https://github.com/meltano/hub/runs/6734716316?check_suite_focus=true

Issue #250, opened 2021-02-24
Add Superset as a plugin


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/hub/-/issues/255

Originally created by @DouweM on 2021-02-24 16:52:39

---

<details><summary>Previous Issue Body</summary>
- https://superset.apache.org/
- https://github.com/apache/superset
- https://pypi.org/project/apache-superset

```bash
meltano add analyzer superset

meltano invoke superset db upgrade # Run automatically?
meltano invoke superset fab create-admin
meltano invoke superset init # Run automatically?
meltano invoke superset run
```

Meltano can manage Superset configuration (https://superset.apache.org/docs/installation/configuring-superset) by allowing values to be set for the keys in https://github.com/apache/superset/blob/master/superset/config.py, automatically generating `superset_config.py`, and pointing Superset there by using the `SUPERSET_CONFIG_PATH` env var. Users should also be able to set `SUPERSET_CONFIG_PATH` (or `meltano config superset set config_path <path>`) themselves to use their own config file.

Ideally, Meltano would also be able to inject database connection strings corresponding to loaders directly into Superset so that these don't need to be managed in two places: https://superset.apache.org/docs/databases/installing-database-drivers, https://superset.apache.org/docs/databases/postgres. Possibly through the `DB_CONNECTION_MUTATOR` setting? https://github.com/apache/superset/issues/9045
</details>

### Meltano 2.0 Scope

Using https://github.com/pnadolny13/meltano_example_implementations/blob/superset_single_container/meltano_projects/jaffle_superset/meltano.yml#L52 (with possibly some https://gitlab.com/meltano/files-superset) we will get Superset to a place where users can install and spin it up and down using Docker. 

- [ ] Entry on the Hub for Superset
- [ ] Possible Files bundle if necessary

MeltyBot:
[View 51 previous comments from the original issue on GitLab](https://gitlab.com/meltano/hub/-/issues/255)

DouweM:
https://github.com/meltano/files-superset/pull/12 is no longer needed now that we have https://github.com/meltano/meltano/pull/6031.

PR #550, opened 2022-06-06
Update transformer, orchestrator, utilities, and files pages to be built from YAML

tayloramurphy:
This PR:

* Updates everything that isn't an Extractor or Loader to have their pages be built from the YAML file
* Moves the Next Steps and Page topic (first sentence) into the ruby generator
  * This can be powered by a new, optional key `definition` key in the YAML
* Adds a "settings preamble" section so we can add more setting context if needed via the YAML
* Updates the JSON schema and API generator
* Removes the previous markdown files

**Netlify Build: https://deploy-preview-550--meltano-hub.netlify.app/**


tayloramurphy:
@DouweM this is the PR that I think will solve your #250 issue with the wrong text at the top. 

I think I'm going to add a new optional key for the plugins so that the Settings section can have a custom set of paragraphs we can insert if we want. The  first sentence on the page and the `Next Steps` section could be in the `meltano_plugin.html` file. I think that should cover what we currently have for the plugins.

tayloramurphy:
@DouweM @pnadolny13 this is ready for review. I updated my first comment with a list of everything that is being updated in here. I recommend looking at the build to get a handle on where the text in the YAML lands.

@DouweM for your superset work, there's a few places where we can add text to the page to make it read well

DouweM:
@tayloramurphy Looks great, just one final bug:

https://deploy-preview-550--meltano-hub.netlify.app/extractors/3plcentral has the correct description for an extractor:

> The tap-3plcentral [Meltano extractor](https://docs.meltano.com/concepts/plugins#extractors) pulls data from [3PLCentral](https://www.3plcentral.com/) that can then be sent to a destination using a [Meltano loader](https://docs.meltano.com/concepts/plugins#loaders).

https://deploy-preview-550--meltano-hub.netlify.app/taps/3plcentral has the INcorrect description for a target, and the "Singer tap" link actually points at https://hub.meltano.com/singer/target.

> The tap-3plcentral [Singer tap](https://hub.meltano.com/singer/target) send data into [3PLCentral](https://www.3plcentral.com/) after it was pulled from a source using a [Singer tap](https://hub.meltano.com/singer/taps).

DouweM (approved):

DouweM:
@pnadolny13 I've merged this, if you had any more suggestions we can pick those up in separate issues/PRs!

PR #568, opened 2022-06-07
refresh static hub metrics files

pnadolny13:
I manually pulled these from S3 so its updated as of today until we fix the CI pipeline


tayloramurphy (approved):
LGTM

PR #569, opened 2022-06-07
Fix loader target_schema extras

DouweM:
After the removal of `env_aliases` (https://github.com/meltano/meltano/pull/5927), `TARGET_[SNOWFLAKE|POSTGRES]_SCHEMA` is only being set when the setting was actually called `schema`, which was the case with the `meltano` variant but not `transferwise` and `datamill-co`, meaning that the extra would always be `null`. This fixes that.


tayloramurphy (approved):
Good catch - thanks @DouweM

aaronsteers (approved):
Great catch. Thanks, @DouweM. 

Cc @kgpayne

PR #583, opened 2022-06-08
Feature: Copy to Clipboard

alexmarple:
https://user-images.githubusercontent.com/54377734/172727553-8c60223b-d2d9-4fd9-93de-349f5cb33029.mov


alexmarple:
@aaronsteers @tayloramurphy i couldn't wait for hack day to do something worthwhile but small. This should be all good to merge. Lemme know if you see something amiss or I somehow missed a spot.

tayloramurphy (changes requested):
Screen Shot 2022-06-13 at 8 44 51 AM

It seems like having a comment in the code block causes the copy link to be super small.

Otherwise it looks good. It makes me think we should split up possible config sections like https://deploy-preview-583--meltano-hub.netlify.app/extractors/tap-auth0#client-secret since the copy gets that entire thing and users wouldn't actually do that.

alexmarple:
@tayloramurphy fixed that bug. I think I'll add a couple of other items to this on hack day but otherwise is a nice, small addition.

tayloramurphy (approved):
Looks like it's fixed! @alexmarple I'm ok to merge as-is so you can have a smaller PR for hack day 👍

PR #587, opened 2022-06-08
Create issue forms for adding plugins to Hub

### LINKED ISSUE #566: Update "Add New Singer Connector" issue template for GitHub
https://github.com/meltano/hub/blob/main/.gitlab/issue_templates/Add%20New%20Singer%20Connector.md is in the `.gitlab` directory, and talks about GitLab features like Merge Requests and the Web IDE. We should update it!

@tayloramurphy For you to schedule

edgarrmondragon:
Closes #566


edgarrmondragon:
This is how it looks. The emoji is only there to prove that it can be added 😄 

Screen Shot 2022-06-08 at 18 02 02
Screen Shot 2022-06-08 at 18 02 21

edgarrmondragon:
@tayloramurphy @pnadolny13 wdyt of this iteration? Is there something you want to see in the issue form?

You can try it out here: https://github.com/edgarrmondragon/hub/issues/new/choose

aaronsteers:
@edgarrmondragon - I'd like to address capabilities here also?

Something like...

```
  - type: checkboxes
    id: capabilities
    attributes:
      label: Tap Capabilities
      description: Please describe the capabilities of this tap. Most taps, including SDK-based taps will support `state` and `catalog` but not `properties`.
      options:
        - label: This tap supports incremental sync, aka `state`
          required: true
          default: true
        - label: This tap supports the `--catalog` input to specify metadata and/or selection logic.
          required: true
          default: true
        - label: This tap supports the `--properties` input (a legacy version of the `--catalog` option)
          required: true
          default: false
        - label: This tap is build on the Singer [SDK](https://sdk.meltano.com) for Taps
          required: true
```

aaronsteers:
UPDATE: An alternative to having users enter plugin capabilities and settings would be to have this in place to get automated confirmation: https://github.com/meltano/hub/issues/603

.github/ISSUE_TEMPLATE/new_tap.yml
```diff
@@ -0,0 +1,40 @@
...
+  id: name
+  attributes:
+    label: Name
+    description: The unique name of the connector
+
+- type: input
+  id: singer_name
+  attributes:
+    label: Singer Name
+    description: The Singer specific name of the connector. Typically tap-<name>
```
tayloramurphy, line 40: @edgarrmondragon I'm ok with this as a first iteration, but can we include some of the steps listed in https://github.com/meltano/hub/blob/main/.gitlab/issue_templates/Add%20New%20Singer%20Connector.md as well? 

I can possibly see that having users enter data in an issue form like this could be a little annoying if it's disconnected from the PR they'd eventually have to create as well, so I'm not sure how far we want to take the YML issue definition

tayloramurphy (changes requested):

.github/ISSUE_TEMPLATE/new_tap.yml
edgarrmondragon: @tayloramurphy yeah, I though the same so I considered doing [something for hackday](https://github.com/orgs/meltano/discussions/388#discussioncomment-2951325) around that using something like https://jsonforms.io/docs/ to create a PR on a user's behalf.

pnadolny13: I like this form idea but yeah I agree with @tayloramurphy that it could end up be annoying if it doesnt help with the PR.

The goal of the previous gitlab template was to basically create an issue with a checklist of things to do. Are we changing the goal to try to allow users to build their yaml/PR using the form? I almost think we should have 2 different templates, or options. One for just creating the issue itself with the todo list when someone already has the yaml definition or plans to build it manually and another for drafting the full yaml via the form for someone whos new and needs to be walked through it because they dont quite know what content is required.

tayloramurphy: ohh @edgarrmondragon that'd be cool! If it's possible to be primarily driven off the JSON schema that'd be amazing so there's still only a SSOT.

tayloramurphy (approved):

PR #590, opened 2022-06-09
Add setting and command to Airflow, fix YAML url

tayloramurphy:


_data/meltano/orchestrators/airflow/apache.yml
```diff
@@ -11,6 +11,9 @@ requires:
   - name: files-airflow
     variant: meltano
 commands:
+  create-admin:
+    args: fab create-admin
```
tayloramurphy, line 15:
```suggestion
    args: create-admin: "users create --username admin --firstname FIRST_NAME --lastname LAST_NAME --role Admin --email admin@example.org"
```

_data/meltano/orchestrators/airflow/apache.yml
```diff
@@ -11,9 +11,32 @@ requires:
   - name: files-airflow
     variant: meltano
 commands:
+  create-admin:
+    args: "users create --username admin --firstname FIRST_NAME --lastname LAST_NAME --role Admin --email admin@example.org"
```
DouweM, line 15: @tayloramurphy This will create the user with first name literally `FIRSTNAME` and last name `LAST_NAME`, and the demo admin. Is that what you intended? I think it'd be better to have it just be `users create` and have the user enter the other details on the prompt.

PR #592, opened 2022-06-11
[fix] Add Missing Env Var to BQ

z3z1ma:
Adds `env` to config for `GOOGLE_APPLICATION_CREDENTIALS` which the tap required.


pnadolny13 (changes requested):
@z3z1ma thanks for the PR! I left a similar comment [on your other PR](https://github.com/meltano/hub/pull/595#pullrequestreview-1004327641), would you mind doing the same here and not updating any formatting for the PR so its easier to see whats actually changing vs just being reformatted?

z3z1ma:
Good to go, probably just squash commit

z3z1ma:
@pnadolny13 
Correct, neither target will work out of the box without that. `env_alias` which was in the old discovery.yml was covering the same thing.

image

pnadolny13 (approved):
@z3z1ma awesome thanks for clarifying for me!

PR #595, opened 2022-06-11
[fix] Incorrect target name and missing env on transferwise bigquery

z3z1ma:
Incorrect target name on transferwise bigquery. 
Side-effect, causes hub page [here](https://hub.meltano.com/loaders/target-bigquery) to have wrong `variant` hard link which take you to a Page Not Found error. 

Also this also requires `GOOGLE_APPLICATION_CREDENTIALS` to be in the env vars when running so I have exposed the config option here which parallels the default variant.

### DIFF
--- a/_data/meltano/loaders/target-bigquery/transferwise.yml
+++ b/_data/meltano/loaders/target-bigquery/transferwise.yml
@@ -11,7 +11,6 @@ capabilities:
 settings:
 - name: credentials_path
   value: $MELTANO_PROJECT_ROOT/client_secrets.json
-  type: string
   label: Credentials Path
   description: |
     Fully qualified path to `client_secrets.json` for your service account.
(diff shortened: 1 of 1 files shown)


pnadolny13 (changes requested):
@z3z1ma good catch on these! Thanks for the PR. Would you mind updating the PR to not do any re-formatting? Its hard to tell whats actually changing in the diffs. I do think its a good idea though, its probably worth having a formatter run over all these files so theyre consistent but maybe in a follow on.

tayloramurphy (approved):
Looks great - merging!

_data/meltano/loaders/target-bigquery/transferwise.yml
```diff
@@ -9,6 +9,17 @@ capabilities:
 - hard-delete
 - soft-delete
 settings:
+- name: credentials_path
+  value: $MELTANO_PROJECT_ROOT/client_secrets.json
+  type: string
```
tayloramurphy, line 14:
```suggestion
```

Issue #597, opened 2022-06-13
CI Validation for Plugin Definition Files


pnadolny13:
This is partially related to the migration to GitHub and also related to the major refactor we did for the 2.0 release but we should fix CI to validate plugin definition files.

- Potentially update current json schemas. Any required changes as part of the refactor.
- I propose that we make label and description required now since most PR review include us asking contributors to add them
- Maybe add a yaml formatter so we're consistent, related to https://github.com/meltano/hub/pull/595#pullrequestreview-1004327641

Side note: I originally wrote the json schemas to validate the meltano API static files because they were being auto generated from the legacy format and were using a new structure (basically what we ultimately migrated all the files to). Theres a possibility that the requirements of the API are not the same as the files. For example we dont need to serve `keywords` and `domain_url` via the API because that metadata added for use in the Hub. It might make sense to have 2 sets of schemas, one for the file definitions and one for the generated API static files with lots of overlap which can be managed in the common/base schemas that are shared.

PR #607, opened 2022-06-15
[feat] Better utility script and set up

z3z1ma:
Added utility to easily spot out all variants with incorrect names. Add pyproject toml so we can run utils more conveniently and manage deps. Updated old jsonschema util to point to correct dirs. 

Fixed lingering tap misname with utility output ;) 

Ran isort and black.

Added name check to CI

### DIFF
--- a/.github/workflows/api_data_validation.yml
+++ b/.github/workflows/api_data_validation.yml
@@ -24,14 +24,18 @@ jobs:
 
     - name: Upgrade pip
       run: |
-        pip install pip
+        pip install pip poetry
         pip --version
+        poetry --version
 
-    - name: Pip installs
+    - name: Poetry install
       run: |
-        pip install ruamel.yaml
-        pip install jsonschema
+        poetry install
 
     - name: Yaml schema validate
       run: |
-        python yaml_validate.py
+        poetry run python yaml_validate.py
+
+    - name: Yaml name validate
+      run: |
+        poetry run python yaml_check_names.py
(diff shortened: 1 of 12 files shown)


aaronsteers:
@pnadolny13 - Can you review this one? Looks good from my side, but I think you and Taylor are closest to this code.

z3z1ma:
So its important to mention as I did in engineering-chatter,

CI has NOT been running jsonschema validation because of incorrect dir in util script. 
It also has a print statement that blows up the logs because of the large number of taps/targets which do not conform to the schema. 

We should drop the print statement for 1️⃣  and hide it behind a verbosity flag

2️⃣  I would also recommend we use pre-commit here. This way we validate users stuff during development?

tayloramurphy:
@z3z1ma thank you for the PR.

@pnadolny13 I'd appreciate a review from you. WRT @z3z1ma's points about the print statement and the pre-commit hook, I think we can address those in separate issues. 

Thoughts on merging even if the validation is failing in this PR and working with a separate issue to get everything passing the validation? Clearly the Hub is "ok" with unvalidated JSON right now, but I don't know if cluttering this PR with all of the fixes is the way to go.

tayloramurphy:
@pnadolny13 sounds good - I've moved #597 to the next iteration. Will you review, approve, and merge this PR if it looks good to your eyes?

pnadolny13 (approved):
@z3z1ma looks good to me! I'll go ahead and merge it in, then I'm assigned to https://github.com/meltano/hub/issues/597#event-6821157228 to clean up the failing schema tests as a follow on.

Issue #599, opened 2022-06-14
429 Response Codes


visch:
To replicate

1. Go to https://hub.meltano.com
1. Click Extractors
1. Click an Extractor
1. Click the MeltanoHub icon on the top left of the page

Do steps 2-3 quickly

Here's a video of this happening

https://user-images.githubusercontent.com/8680264/173482873-a61af4d0-c427-4411-9e8c-82003b30a17b.mp4

--

Wild guess

Something to do with all the requests for Images and no cache control on this images?

tayloramurphy:
Ha! Yeah - can confirm I get the same thing. Likely a limitation on Netlify's end. We have https://github.com/meltano/internal-general/issues/247 which is an internal issue about moving hosting to Cloudflare. 

I'm not sure there's much we can do here. @aaronsteers any ideas?

pnadolny13:
@tayloramurphy @alexmarple I also ran into this today and got some weird buggy side affects. Is there any short term fixes we can do on our side to cache those images or potentially load only a subset? It shows up for me almost immediately so the browsing experience is fairly interrupted.

alexmarple:
@pnadolny13 oh that's fun. we could try a plugin like: https://github.com/wildlyinaccurate/jekyll-responsive-image with cacheing set to true. this might definitely be an issue with netlify but i feel like they wouldn't do us dirty like that.

pnadolny13:
@tayloramurphy I'd consider this a relatively high urgency bug, its a struggle to use the hub in its current state. If I dont land on what im looking for quickly then I start to get lots of 429's so I have to stop for a minute and let it cool down 😄 . Do we think Cloudflare will happen soon and do we know if it would resolve this issue?

@alexmarple is that something thats easy to test out? I'd assume that making the change PR then testing on the preview app would give us a good sense of whether it helps while still using Netlify or not.

pandemicsyn:
@alexmarple and @pandemicsyn - What likely root causes would drive this and what mitigations would be helpful? Could this be related to the new omnipresent search box backend needing to loading all images on all pages?

429 means netlify is rate limiting the user , so caching would definitely help. Just loading `https://hub.meltano.com/extractors/` results in **355** (🤯 ) requests! Thats gonna be really iffy experience for users with slower connections.

visch:
Cache-Control headers have to go somewhere, webserver is going to do it wherever that's at.

[...]

PR #619, opened 2022-06-22
Basic cache control from _headers file

pandemicsyn:
Quick example of how to set cache control headers, the example includes all files in `/assets/logos/*` but we might want to include others. This should cover all the logos though which is bulk of the problem at the moment:

Screen Shot 2022-06-22 at 2 09 37 PM

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -63,4 +63,4 @@ highlighter: rouge
 plugins:
   - jekyll-sitemap
 
-include: [_redirects]
+include: [_redirects, _headers]
--- a/_headers
+++ b/_headers
@@ -1,0 +1,3 @@
+/assets/logos/*
+    cache-control: max-age=604800
+    cache-control: public


tayloramurphy:
I'm okay to merge this but it sounds like we'll just have to be smarter on this for the hub redesign. 

@aaronsteers or @alexmarple I'd like a review from you to give code approval.

aaronsteers (approved):
Thanks, @pandemicsyn. While I couldn't repro the issue consistently enough to confirm this as a fix for #599, the caching makes sense and I don't see any reason to _not_ merge. (Preview build running smoothly on Netlify with no detectable issues.)

Since today is F+F I'll go ahead and merge this, but will keep #599 open with a request for confirmation of fix before closing. Thanks!

cc @tayloramurphy, @pandemicsyn

PR #623, opened 2022-06-23
Add extra dbt and sqlfluff commands

### LINKED ISSUE #621: Add the dbt build command to the dbt transformer plugin
Currently all dbt commands are there, however the command responding to `dbt build` is missing. 

Should this be added [here](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/core/bundle/discovery.yml#L2985)? If that is the case, I can create a pull request!

JulesHuisman:
Closes #621 

This adds some extra commands for DBT and adds the `sqlfluff fix` command to the sqlfluff utility.

I was not sure if I was supposed to pull directly into main, but let me know.


tayloramurphy (approved):
Thanks for this PR! Looks good to me. @pnadolny13 would you give it another quick once over and then merge if you approve?

PR #627, opened 2022-06-24
Update transferwise to have properties capability

tayloramurphy:
Based on https://github.com/transferwise/pipelinewise-tap-postgres#run-the-tap-in-sync-mode


tayloramurphy:
Closing as it seems the tap supports both 🙃 https://github.com/transferwise/pipelinewise-tap-postgres/blob/14a0be55e8494f7f5b7ee4e6627ece902db6729a/tap_postgres/__init__.py#L380-L382

PR #629, opened 2022-06-26
Add PR template

WillDaSilva:
Seems like it's only practical to have a single PR template per project. Technically you can have multiple, but GitHub does not provide any UI to have them show up in that case. If you have multiple PR templates, then users would have to add the `?template=some_pr_template.md` to their "create PR" URL.

In light of this, I have merged the two MR templates from GitLab to create the new one for GitHub.

I have also deleted the `.gitlab` directory as everything from it has now been migrated to GitHub.


aaronsteers (approved):
Thanks, @WillDaSilva ! A good iteration.

tayloramurphy (approved):

Issue #223, opened 2022-04-05
Revise directory structures, move python files into sensible subdirectories.


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/hub/-/issues/227

Originally created by @aaronsteers on 2022-04-05 21:04:09

---

We have a ton of bare python scripts at the root directory of the repo. I recommend we put in subdirectories according to function, perhaps with a short readme in each subdirectory that explains the grouping of those scripts together.

MeltyBot:
[View 4 previous comments from the original issue on GitLab](https://gitlab.com/meltano/hub/-/issues/227)

pnadolny13:
This looks to have been completed!

PR #630, opened 2022-06-27
Fix ci schema validation

pnadolny13:
Closes https://github.com/meltano/hub/issues/597 and https://github.com/meltano/hub/issues/223

The core changes in this PR are to clean up the json schemas following everything that happened during the 2.0 release. From my view theres 2 things we want to do during CI:

1. Plugin Definition YAML Validation: new additions and updated plugin definition yaml files need to meet our schemas. Users should get failing tests vs us trying to catch definition bugs in the PR review.
2. API Static File Validation:  the generated API static files need to be validated against the expectations of Meltano in terms of lock file plugin definitions (i.e. we shouldnt send hub specific metadata via the API). No changes to the plugin definition schemas (i.e. adding a key but forgetting to exclude it from the API) or Ruby code changes for generating the API static files should break our contract with Meltano. 

To do this I took the schemas that we had previously, which were intended to validate the API files but were morphed slightly during the 2.0 push, and split them out into two sets of schemas with a large overlap. The intention is to have 1 set for asserting the autogenerated API files and another set for asserting the hub plugin definition files that live in the repo itself. Theres a huge amount of overlap which I was able to pull out into a `common` schema directory. Its slightly complex but it allows us to be more precise and defining the schemas separately. If later we decide to rejoin them and [pull the hub metadata out](https://github.com/meltano/hub/discussions/609) then we can easily do that but for now we can be comfortable that everything meets our expectations.

CI:
- Rename api_data_validation.yml to plugin_definition_validation.yml since its really running validation on the yaml files not the API files. Jekyll test workflow asserts the API files because it needs to run a build before asserting.
- Move jekyll test to use poetry for python related stuff like our other workflow already does.

Scripts:
- moved all top level scripts into a `utility_scripts` directory split out by `api` and `plugin_definitions`. I wasnt totally sure what some of these scripts are for and I know we committed a few utilities that were just for running one off operations so I defaulted to just keeping them for now.

Schemas:
- split the schemas into `plugin_definitions`, `api`,  and `common`. The common schemas are referenced in the others. Mostly the plugin_definitions/api schemas are for defining the top level keys that are allowed so that we can set `"additionalProperties": false,` and make sure no extra keys or mispellings are included. The other type specific and common schemas are for actually asserting the data types, content, etc. 
- the api directory has additional schemas for asserting the index APIs
- the plugin_definition directory  has an additional schema for asserting the hub metadata fields like `maintenance_status`, `usage` markdown, etc.

### DIFF
--- a/schemas/common/base_setting.schema.json
+++ b/schemas/common/base_setting.schema.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "base_setting.schema.json",
+    "$id": "common/common/base_setting.schema.json",
     "title": "JSON Schema for base_settings",
     "description": "Meltano (https://meltano.com) is an open source platform for building, running & orchestrating ELT pipelines",
     "type": "object",
--- a/schemas/common/common.schema.json
+++ b/schemas/common/common.schema.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "common.schema.json",
+    "$id": "common/common.schema.json",
     "title": "JSON Schema for common plugin properties.",
     "description": "Meltano (https://meltano.com) is an open source platform for building, running & orchestrating ELT pipelines",
     "type": "object",
--- a/schemas/common/extractors_specific.schema.json
+++ b/schemas/common/extractors_specific.schema.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "extractors_specific.schema.json",
+    "$id": "common/extractors_specific.schema.json",
     "title": "JSON Schema for extractor plugins.",
     "description": "https://docs.meltano.com/concepts/plugins#extractors",
     "type": "object",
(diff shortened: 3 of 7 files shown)


schemas/plugin_definitions/loaders.schema.json
```diff
@@ -0,0 +1,49 @@
...
+    "$id": "loaders.schema.json",
+    "title": "JSON Schema for loader plugins.",
+    "description": "https://docs.meltano.com/concepts/plugins#loaders",
+    "type": "object",
+    "allOf": [
+        {
+            "$ref": "common.schema.json"
+        },
+        {
+            "$ref": "loaders_specific.schema.json"
```
rwfeather, line 12: I think if we want the YAML language server in IDEs to be able to pick these up and understand them, we need to specify the relative filepath

image

Updating these to be `../common/common.schema.json` and `../common/loaders_specific.schema.json` will validate the plugin definitions inside a users IDE (if they have it set up to do so):

image

Might be valuable to improve the authoring experience (and maybe catch problems before CI even gets to it)

rwfeather:
@pnadolny13 The schema organization is amazing! This will definitely be flexible enough to let the API and Definition schemas evolve independently.

The video walkthrough was very helpful, I feel confident that engineering could maintain this going forward.

rwfeather (approved):

pnadolny13:
@rwfeather I'm going to go ahead and merge this in! Thanks for the review

PR #633, opened 2022-06-28
add more next steps to sqlfluff

pnadolny13:
I've used this approach in squared to make SQLFluff export my required dbt env vars so they look like theyre coming from dbt https://github.com/meltano/squared/blob/cdbfc6a15ae6dd9e6c3f3b4baca94b0159ddcd1b/data/utilities/utilities.meltano.yml#L3.

@edgarrmondragon we were just talking in slack about this, do you think this is a good approach to recommend? Any issues with editing the namespace?

### DIFF
--- a/_data/meltano/utilities/sqlfluff/sqlfluff.yml
+++ b/_data/meltano/utilities/sqlfluff/sqlfluff.yml
@@ -23,4 +23,4 @@ next_steps: |
   1. Run `meltano invoke sqlfluff:lint` to lint your SQL files.
   1. Optionally, if you're using dbt and your project is expecting environment variables to be set like `DBT_SNOWFLAKE_ACCOUNT`
     you can set the sqlfluff `namespace` setting in your meltano.yml to `dbt_<adapter_name>` (using your specific adapter i.e. `dbt_snowflake`)
-    so SQLFluff's settings are exported with the `DBT_<ADAPTER_NAME>` prefix properly.
+    and configure any required dbt setting under your SQLFluff utility so they're exported with the `DBT_<ADAPTER_NAME>` prefix properly.
--- a/_data/meltano/utilities/sqlfluff/sqlfluff.yml
+++ b/_data/meltano/utilities/sqlfluff/sqlfluff.yml
@@ -23,4 +23,4 @@ next_steps: |
   1. Run `meltano invoke sqlfluff:lint` to lint your SQL files.
   1. Optionally, if you're using dbt and your project is expecting environment variables to be set like `DBT_SNOWFLAKE_ACCOUNT`
     you can set the sqlfluff `namespace` setting in your meltano.yml to `dbt_<adapter_name>` (using your specific adapter i.e. `dbt_snowflake`)
-    and configure any required dbt setting under your SQLFluff utility so they're exported with the `DBT_<ADAPTER_NAME>` prefix properly.
+    and configure any required dbt setting under your SQLFluff utility so they're exported with the `DBT_<ADAPTER_NAME>` prefix properly when SQLFluff runs.
--- a/_data/meltano/utilities/sqlfluff/sqlfluff.yml
+++ b/_data/meltano/utilities/sqlfluff/sqlfluff.yml
@@ -19,4 +19,8 @@ logo_url: /assets/logos/utilities/sqlfluff.png
 definition: is a linting tool for SQL files, often used with dbt to enforce SQL code standards.
 next_steps: |
   1. Customize [the config files](https://docs.sqlfluff.com/en/stable/configuration.html) to match your team's style guide.
+    SQLFluff comes with logic defaults so you will only need to override configurations you want to change.
   1. Run `meltano invoke sqlfluff:lint` to lint your SQL files.
+  1. Optionally, if you're using dbt and your project is expecting environment variables to be set like `DBT_SNOWFLAKE_ACCOUNT`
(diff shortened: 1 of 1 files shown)


edgarrmondragon:
> what functionality do we lose by considering it a custom plugin? I'm guessing `meltano lock` would stop working, maybe other stuff too?

@pnadolny13 Yes, `meltano lock` would no longer work, and in general all settings and commands would have to be rewritten in `meltano.yml` since a lock file would not be used.

pnadolny13:
> @pnadolny13 Yes, `meltano lock` would no longer work, and in general all settings and commands would have to be rewritten in `meltano.yml` since a lock file would not be used.

Hmm thats probably not ideal then, can you think of a better way to do this? Should the functionality youre talking about not rely on namespace in this way? I'm not sure if this is a SQLFluff only use case or if its worth finding a pattern that generally works. I guess overriding with `env` keys to prefix like `DBT_SNOWFLAKE_X` also works but it leaves a fair bit of manual configuration for SQLFluff users before its functional. It makes me think we should have adapter specific sqlfluff utilities with all the settings pre-defined, thats pretty messy though.

edgarrmondragon:
> Hmm thats probably not ideal then, can you think of a better way to do this? Should the functionality youre talking about not rely on namespace in this way? I'm not sure if this is a SQLFluff only use case or if its worth finding a pattern that generally works. I guess overriding with `env` keys to prefix like `DBT_SNOWFLAKE_X` also works but it leaves a fair bit of manual configuration for SQLFluff users before its functional. It makes me think we should have adapter specific sqlfluff utilities with all the settings pre-defined, thats pretty messy though.

@pnadolny13 These are the current options AFAICT:

|| Downside |
|-|-|
| Override `namespace` to `dbt_<adapter>` to reuse the `dbt-<adapter>` settings. | SQLFluff becomes a _custom_ plugin and `commands` have to be redefined in `meltano.yml`. |
| Copy the `dbt-<adapter>` setting definitions into the `sqlfluff` utility in `meltano.yml`. The lockfile is still used for `commands`, etc. | Setting definition duplication. |
| Manual configuration with environment variables, i.e. `DBT_<ADAPTER>_<SETTING>` | Defeats the purpose of Meltano's multiple configuration layers. |

I wonder if Meltano should have a concept of _add-on_ plugins that extend the functionality of other plugins, and whose settings (and even `venv`!, see https://github.com/meltano/meltano/issues/3000) can be reused.

cc @tayloramurphy

pnadolny13:
@tayloramurphy I'm looping back around on this and wanted to get your take. I'd really like to get sqlfluff in a good place for users to easy install, at least with the common dbt+warehouses combos.

Edgar laid out our options really nicely above in https://github.com/meltano/hub/pull/633#issuecomment-1169118424. 

Maybe we want Meltano to have a way supporting this internally like described in:

>I wonder if Meltano should have a concept of add-on plugins that extend the functionality of other plugins, and whose settings (and even venv!, see https://github.com/meltano/meltano/issues/3000) can be reused.

But even for the short term it would be great to have something that users can install in a semi-clean fashion today. I think either adding pre-configured warehouse adapters like `sqlfluff-snowflake` or writing a big blob of instructions on the plugin page that lays out the setting you need to manually configure to get sqlfluff working with your dbt plugin.

Here the squared example of how I have it working https://github.com/meltano/squared/blob/9bf062e22b95cf547c5f6692c3edf7a61629ac46/data/utilities/utilities.meltano.yml#L3 without overriding the namespace which would be cleaner but make it become a custom plugin with no lock file.

tayloramurphy:
@pnadolny13 I think looking at https://github.com/meltano/meltano/issues/3000 is a better long-term solution. I'm in favor right now of `writing a big blob of instructions on the plugin page` for sqlfluff.

pnadolny13:
> @pnadolny13 I think looking at [meltano/meltano#3000](https://github.com/meltano/meltano/issues/3000) is a better long-term solution. I'm in favor right now of `writing a big blob of instructions on the plugin page` for sqlfluff.

@tayloramurphy that's a fine short term fix. I can add a snowflake example blob pretty easily.

I would clarify though that I think that issue was more focused around the executables and not the settings/configs like I'm more concerned with here. That issue would need to share a venv AND all of its settings with sqlfluff. I'm unsure of whether this is a common enough pattern to build something into meltano core for it.

vjanssens:
As discussed yesterday through Slack and what you mentioned in #726, i'd like to see the following added to the sqlfluff docs when using DBT:

1. Let users know they need to add specific packages to the `pip_url` setting. I've had to manually add these: `dbt-core==1.2.0 dbt-postgres==1.2.0 sqlfluff-templater-dbt==1.2.1`

Like: 
```
utilities:
    - name: sqlfluff
      variant: sqlfluff
      pip_url: dbt-core==1.2.0 dbt-postgres==1.2.0 sqlfluff[dbt]==1.2.1 sqlfluff-templater-dbt==1.2.1
      executable: sqlfluff
```

2. You need a .sqlfluff file in the root of your project to get it to work, something along the lines of:

```
[sqlfluff]
templater = dbt
dialect = postgres

[sqlfluff:templater:dbt]
project_dir = transform
profiles_dir = transform/profile
profile = <profile name>
```

3. And a .sqlfluffignore file, like:
```
.meltano/
utilities/
transform/dbt_packages/
target/
dbt_modules/
macros/
```

pnadolny13:
@edgarrmondragon @tayloramurphy @rwfeather @vjanssens I updated this based on your feedback, including a Snowflake example. Let me know what you think.

Its still not perfect :unamused: because the user needs to figure out how to mock their own dbt setting but at least its something to help them in the current implementation.

rwfeather (approved):
lgtm, nice documentation

PR #641, opened 2022-07-05
Update stats language, telling users that they include all variants

### LINKED ISSUE #639: Re-label metrics to make it clear that theyre not variant level
The metrics shown on the plugin page includes all variants combined. Its documented in https://hub.meltano.com/singer/docs#meltano-usage-metrics but its a little mis leading on the plugin page.

Since we have another iteration to add variant level metrics, it would be good to relabel the current metrics for now. Maybe 
"Executions (Last 3 Months)" -> "All Variant Executions (Last 3 Months)" or renaming the "Meltano Stats" header to reflect that its all variants. Later we can add variant specific metrics and probably remove the old non-variant specific metrics.

cc @rwfeather @tayloramurphy @alexmarple

rwfeather:
Closes #639 

Wanted to avoid adding too many words to the badge itself, hopefully this fits the requirements.


rwfeather:
Screen Shot 2022-07-05 at 3 05 21 PM

PR #642, opened 2022-07-05
Add action for daily rebuilds, grab new metrics

### LINKED ISSUE #640: Rebuild MeltanoHub daily to get fresh metrics
Now that we're pulling metrics into the Hub via Netlify when building, we should rebuild this on a daily basis to get up to date metrics with the work done in the Squared project. 

@rwfeather suggested doing something with GitHub Actions if there isn't a way to natively rebuild the site on a daily basis in Netlify.

rwfeather:
Closes #640 

This action will be run 15 minutes after the updated plugin metrics are pushed to s3. Triggers a netlify rebuild via hook.
Will get us back to having fresh metrics every day

CC @tayloramurphy @aaronsteers

### DIFF
--- a/.github/workflows/daily_metrics_rebuild.yml
+++ b/.github/workflows/daily_metrics_rebuild.yml
@@ -3,6 +3,7 @@ name: nightly-netlify-build
 on:
   schedule:
     - cron: "15 9 * * *"
+  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/daily_metrics_rebuild.yml
+++ b/.github/workflows/daily_metrics_rebuild.yml
@@ -1,0 +1,14 @@
+name: nightly-netlify-build
+
+on:
+  schedule:
+    - cron: "15 9 * * *"
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/daily_metrics_rebuild.yml
```diff
@@ -0,0 +1,14 @@
+name: nightly-netlify-build
+
+on:
+  schedule:
+    - cron: "15 9 * * *"
+
```
pnadolny13, line 6: @rwfeather what do you think about adding `workflow_dispatch` so its possible to [manually trigger](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) this workflow? There might be cases where the upstream fails so it would be helpful to be able to manually rerun this if needed.

rwfeather: Good idea, I've updated the action to include it

pnadolny13 (approved):
@rwfeather Looks good to me!

PR #659, opened 2022-07-07
Pull data from top level dir, discard copy

rwfeather:
This points the gridsome server at the top level dir `_data/meltano` instead of the copy. Depending on how long this branch lives, maintaining two copies will likely be difficult.


alexmarple (approved):
LGTM

PR #664, opened 2022-07-11
Add target-miso in loaders

simonpai:
Hi @pnadolny13,
I would like to add a loader entry to the hub. Thanks for your effort in advance!


_data/meltano/loaders/target-miso/miso.yml
```diff
@@ -0,0 +1,34 @@
+description: Personalized Search & Recommendation Engine
+label: Miso
+name: target-miso
+logo_url: /assets/logos/loaders/miso.png
+namespace: target_miso
+variant: miso
```
rwfeather, line 6: The variant name for a meltano plugin is generally going to be the same as the github organization name in the `repo` url. In which case you'll want to update this to be `misoai`.

(@pnadolny13 let me know if I'm mistaken or if its not a hard rule)

pnadolny13: @rwfeather yeah thats been the convention. I think we use the variant name when build the plugin page badge urls so those wont work right now. I think we should plan to decouple those things though in the future.

simonpai: Updated!

rwfeather (approved):
lgtm, thank you

PR #671, opened 2022-07-14
Create pages for default plugins without variant

### LINKED ISSUE #655: Revamp: Setup redirects for subfolder plugin names to default variant

rwfeather:
Closes #655 

Rather than trying to create and manage redirects, I've instead set up some dynamic page generation that creates pages at
`/:type/:name` for each plugin which renders the same and uses the same data as `/:type/:name/:variant` when variant is the default.

You can see here after building the site that we end up with a new html file one level higher than the others:
Screen Shot 2022-07-14 at 12 20 02 PM

It makes the site bundle slightly bigger but not by much.

Check it out here:

https://deploy-preview-671--meltano-hub-gridsome.netlify.app/extractors/tap-amazon-ads-dsp/

is the same as

https://deploy-preview-671--meltano-hub-gridsome.netlify.app/extractors/tap-amazon-ads-dsp/singer-io/

### DIFF
--- a/hub-2.0/ui/gridsome.server.js
+++ b/hub-2.0/ui/gridsome.server.js
@@ -11,7 +11,7 @@ const path = require("path");
 
 const dataRoot = "../../_data/";
 
-const defaultVariants = yaml.load(
+const defaultVariantData = yaml.load(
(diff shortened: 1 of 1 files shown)


alexmarple (approved):
LGTM!

PR #672, opened 2022-07-14
add missing maintainers and assert during CI

pnadolny13:
I noticed a bunch of missing maintainers so I update the existing script to find them and add them. It also exits with code 1 if any are missing or extra. So it can help find and add maintainers while making changes but also validate in CI that they all exist and theres no extras.

- It also reordered the current file alphabetically so some shifting happened, a one time updates
- Add to github test pipeline


rwfeather (approved):
Looks great!

PR #673, opened 2022-07-15
target-csv hotglue, remove version pin

visch:
Version pin caused a Windows failure for me. I like to pin, but on my own terms :D

### DIFF
--- a/_data/meltano/loaders/target-csv/hotgluexyz.yml
+++ b/_data/meltano/loaders/target-csv/hotgluexyz.yml
@@ -5,7 +5,7 @@ namespace: target_csv
 logo_url: /assets/logos/loaders/csv.png
 variant: hotgluexyz
 repo: https://github.com/hotgluexyz/target-csv
-pip_url: git+https://github.com/hotgluexyz/target-csv.git@0.3.3
+pip_url: git+https://github.com/hotgluexyz/target-csv.git
 settings:
 - name: destination_path
   label: Destination Path


edgarrmondragon:
> Version pin caused a Windows failure for me

@visch what's the specific failure you saw? Cause it might be a bug in Meltano if git refs can't be used in pip URLs on Windows 😬.

> I like to pin, but on my own terms

You can still override `pip_url` in `meltano.yml` AFAICT. I'm not sure where we're gonna land on pinning versions in the Hub vs resolving/pinning them when a lockfile is generated.

visch:
> > Version pin caused a Windows failure for me
> 
> @visch what's the specific failure you saw? Cause it might be a bug in Meltano if git refs can't be used in pip URLs on Windows 😬.
> 
> > I like to pin, but on my own terms
> 
> You can still override `pip_url` in `meltano.yml` AFAICT. I'm not sure where we're gonna land on pinning versions in the Hub vs resolving/pinning them when a lockfile is generated.

Sorry for the lack of detail, late night!

Specifically it was this issue https://github.com/hotgluexyz/target-csv/issues/1 which their team fixed in a seperate commit, just a tap related issue. We could bump the version here but based on the other taps/targets we're not doing that for others?

aaronsteers:
This is admittedly a hard problem. Discussion started here: https://github.com/meltano/meltano/discussions/6445

pnadolny13:
@aaronsteers @edgarrmondragon @rwfeather how do we feel about merging this?

I'm with Derek that, at least in the short term until we have a pinning strategy, we should unpin and recommend users pin in their own project. The majority of the plugins on the hub right now are already unpinned, I only see 3 that are pinned right now and one of them ended up causing this bug https://github.com/meltano/hub/issues/708.

pnadolny13 (approved):
@visch this looks good to me! Sounds like we'll go towards not pinning until we have a better strategy for maintaining versions.

PR #694, opened 2022-07-29
Fix lint and format

rwfeather:
Formatter wasn't running correctly due to an HTML syntax weirdness.

Running `yarn format:write` gave a bunch of these:

```
src/templates/Utilities.vue
[error] src/templates/Utilities.vue: SyntaxError: Unexpected closing tag "p". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags (50:11)
[error]   48 |           <p>
[error]   49 |             The current capabilities for <pre class="inline-code-block"><code>{{ $page.utilities.name }}</code></pre> may have been automatically set when originally added to the Hub. Please review the capabilities when using this extractor. If you find they are out of date, please consider updating them by making a pull request to the YAML file that defines the capabilities for this extractor.
[error] > 50 |           </p>
[error]      |           ^^^^
[error]   51 |           <p>This plugin has the following capabilities:</p>
[error]   52 |           <ul>
[error]   53 |             <li v-for="(capability, index) in $page.utilities.capabilities" v-bind:key="index">{{ capability }}</li>
error Command failed with exit code 2.
```

Swapping out a few `<p>` tags for `<div>` tags seems to have resolved it, letting prettier touch those files again.

This PR also fixes a few lingering eslint errors in `gridsome.server.js`


alexmarple (approved):
👍

Issue #689, opened 2022-07-29
Ensure `state` capability is included for all that support it


pnadolny13:
Related to this thread in slack https://meltano.slack.com/archives/C01TCRBBJD7/p1659124262172409 and https://github.com/meltano/hub/pull/688.

I've heard users get tripped up by this a bunch of times where our plugin definition doesnt have `state` as a capability so it does a full refresh every time. Then usual path is that the user looks up what singer state is and starts trying to figure out how to store it and pass it into their commands when really the problem is that meltano is unaware of the fact that it should be doing that automatically. This is a pretty painful flow for new users.

We should take a pass at making sure every tap that doesnt have state as a capability actually doesnt have state. It might be a small number of them so solving this could mean writing a quick script to find taps without state and opening up their repos to see if they should get it added. If its a lot we might have to find a different non-manual route.

tayloramurphy:
@pnadolny13 I'm a fan of it. Likely prior to the Hub revamp launch would be a good time. Ideally we could do something automated as well..

pnadolny13:
I realized that now that we have the hub definitions in Snowflake this will be very easy to query for, and we can choose variants with the highest usage first.

PR #725, opened 2022-08-15
Add missing state capability for extractors

pnadolny13:
Closes https://github.com/meltano/hub/issues/689

Update all extractor variants that are missing the state capability. I wrote a script to iterate all extractor variants that dont have the state capability and pull their README from github to see if they have any of these strings `['state.json', 'incremental', 'state file', '--state']`.

As part of that I found `tap-call-cap` was removed but still has a default variant set, so I removed it. I also found 11 variant repos that are no longer available on github. I'll open a separate PR to remove those.

### DIFF
--- a/_data/default_variants.yml
+++ b/_data/default_variants.yml
@@ -77,7 +77,6 @@ extractors:
   tap-brightpearl: zookal
   tap-bronto: fishtown-analytics
   tap-bynder: radico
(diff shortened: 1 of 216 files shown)


tayloramurphy (approved):
Epic! @pnadolny13 it may be worth running this as part of CI every once in a while too.

PR #727, opened 2022-08-16
add setting to transferwise tap-mongodb and set as default

pnadolny13:
Based on a thread in slack https://meltano.slack.com/archives/C01TCRBBJD7/p1660571129770879?thread_ts=1660402531.034259&cid=C01TCRBBJD7

The context is that I asked if transferwise should be the default over singer-io. I also see more successful telemetry usage for transferwise:

>I'm curious if you evaluated the singer-io variant vs transferwise while you were working on this. Right now singer-io is the default variant but I've heard a fair amount of community members choosing the transferwise variant, I'd like to make that the recommended default if its the best option. Do you have any insight you can share? What were the reasons you chose transferwise over singer-io?

And the response was:

>hi, yes, the default one dosen’t support some important options for mongodb authentication
>one example is the usage of mongodb+srv
>also a authSource different than the database you want to get data
>that’s quite normal for any mongodb deployment
>usually auth goes into a admin database and you use another for storing data

cc @sjcotto

### DIFF
--- a/_data/default_variants.yml
+++ b/_data/default_variants.yml
@@ -201,7 +201,7 @@ extractors:
   tap-mixpanel: hotgluexyz
   tap-mode: franloza
   tap-monday: gthesheep
-  tap-mongodb: singer-io
+  tap-mongodb: transferwise
   tap-ms-teams: singer-io
   tap-mssql: wintersrd
   tap-mysql: transferwise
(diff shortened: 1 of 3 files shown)


pnadolny13:
I also saw a thread in slack where @z3z1ma was talking about his new SDK based variant https://github.com/z3z1ma/tap-mongodb. If thats ready that could be considered the default instead? It just sounds like singer-io wasnt doing what users needed.

tayloramurphy (approved):

tayloramurphy:
@pnadolny13 I'm okay to change it to transferwise to start and then we can evaluate the SDK based one later.

PR #728, opened 2022-08-16
remove where repos are missing

pnadolny13:
As part of https://github.com/meltano/hub/pull/725 I was pulling data from each github repo and noticed that some of them have been deleted or maybe they just made them private but either way they arent accessible anymore.


tayloramurphy (changes requested):
@pnadolny13 it looks like there is a singer-io variant for tap-apprise. Can you review the others to make sure there isn't a quick replacement we could put in its place?

pnadolny13:
> @pnadolny13 it looks like there is a singer-io variant for tap-appfigures. Can you review the others to make sure there isn't a quick replacement we could put in its place?

@tayloramurphy good catch! I didnt see that repo. I found one other tap-outreach that I also put back using the singer-io variant.

tayloramurphy (approved):

PR #731, opened 2022-08-17
Delete Jekyll and move gridsome to the root

alexmarple:
This PR deletes the Hub's Jekyll setup and keeps the Gridsome setup.


_redirects
```diff
@@ -1,69 +0,0 @@
-/taps/sage-intacct /taps/intacct
```
rwfeather, line 1: Do we know if anything was relying on these redirects? The `_redirects` file is picked up by Netlify and should still work after we get rid of jekyll.

_headers
```diff
@@ -1,3 +0,0 @@
-/assets/logos/*
```
rwfeather, line 1: The `_headers` file is picked up by Netlify and should still work after we get rid of jekyll. I don't know if this cache control stuff is necessary though

rwfeather (approved):
Pointed out a couple files I think we should keep but everything else looks like it can be tossed.

rwfeather:
I think the netlify build failure is because this originally targeted `main`. Can't get it to rerun so I'm gonna see if merging `gridsome-main` back into this branch does anything

rwfeather:
@alexmarple @aaronsteers This PR is ready to go and I did my best to keep the change scope as focused as possible.

I suggest if you want to review, look at the commits individually rather than the whole set. I've tried to keep each commit focused on a single step.

- [ ] The netlify build failures will need to be resolved just prior to/just after merging because the gridsome preview site (https://meltano-hub-gridsome.netlify.app/) has it's build config set up via UI:

Screen Shot 2022-08-24 at 12 27 16 PM

- [ ] A follow up PR will be immediately created that deletes the `hub-2.0/ui/src/assets/logos` directory (and only that). This PR already switches us over to the correct directory, I just didn't want to balloon the change set any more than I already did.

_redirects
aaronsteers: cc @alexmarple, @afolson, @pandemicsyn, @tayloramurphy 

Quick check-in on whether we need these redirects? I imagine yes for safety and to not break existing links? 

@rwfeather - How difficult is it to add these redirects also for Gridsome/Vue? I don't think we should block this PR but  if needed, we can create a follow-on issue to add those back before merging `gridsome-main` to `main`.

_headers
aaronsteers: @rwfeather - Apologies that I'm not super well versed in the implications. Do we need to do anything more to address, in this PR or in a follow-on? If a follow-on is warranted, can you log an issue for that action? Thanks!

aaronsteers:
@rwfeather re:

> @alexmarple @aaronsteers This PR is ready to go and I did my best to keep the change scope as focused as possible.

Fantastic! Thanks for the quick work on this.

> I suggest if you want to review, look at the commits individually rather than the whole set. I've tried to keep each commit focused on a single step.

👍 

> * The netlify build failures will need to be resolved just prior to/just after merging because the gridsome preview site ([meltano-hub-gridsome.netlify.app](https://meltano-hub-gridsome.netlify.app/)) has its build config set up via UI:

Understood 👍 

> * A follow up PR will be immediately created that deletes the `hub-2.0/ui/src/assets/logos` directory (and only that). This PR already switches us over to the correct directory, I just didn't want to balloon the change set any more than I already did.

Sounds good!

_redirects
tayloramurphy: @aaronsteers yes, we need those redirects still. I don't know what would ever cause us to not need them honestly. But we added those fairly recently so we should keep them.

rwfeather: @aaronsteers I could have been a little clearer. All we need to do in this case is keep the `_redirects` file (it is a Netlify config and will keep working after the switch). Comment was made when I noticed that it was marked for deletion in this PR.

@tayloramurphy Thanks for clarifying

I'll revert the deletion.

aaronsteers:
@rwfeather - I see the latest merge has created conflicts. Once those are resolved, I'd be supportive of merging even if checks are not 100% passing. We can continue to debug those over in the `gridsome-main` branch, and merging sooner than later will help minimize extra work from future conflicts.

aaronsteers (approved):

_headers
alexmarple: @rwfeather what AJ said? I think you're just calling this out but that sounds fine to me.

rwfeather: I made these comments when I was reviewing it as Alex's PR. Just some files I noticed we might not want to delete. I'll be adding a commit that reverts that.

PR #735, opened 2022-08-19
change: keywords as single line with wordwrap

aaronsteers:
When there are a lot of keywords (as with Spreadsheets Anywhere), this makes a more condensed output.

image

Note: In future, it might be night to have each of these to hyperlink back to a search for that keyword. (Above my capabilities as of now. 😅 )


alexmarple (approved):
👍

PR #737, opened 2022-08-19
add incremental replication option to target-bigquery

tayloramurphy:
Added language from https://github.com/adswerve/target-bigquery#target-config-file and adds the missing option.


pnadolny13 (approved):
@tayloramurphy looks good to me

PR #738, opened 2022-08-22
Add variant links to plugin page

### LINKED ISSUE #705: Revamp: Add variant alternatives to plugin pages
We need to point out other variants that are available on template pages.

rwfeather:
Closes #705 

This populates a new section at the top of the plugin page, showing what variants are available, which one is the default, and which one the user is currently on.

The entry in the list for the current page will not be a link to indicate the user is already looking at that variant, but until we fix the link styling it won't be apparent.


alexmarple (approved):
👍

PR #746, opened 2022-08-24
Put netlify config files in right dir to get picked up by gridsome build

rwfeather:
Noticed this as I was keeping an eye on the post-merge build:

Screen Shot 2022-08-24 at 4 15 16 PM

Jekyll was picking these up from the root directory and including them in the bundle. In order for Gridsome to do that, we need to have them in a folder called `static`


alexmarple (approved):
👍

PR #749, opened 2022-08-25
Swap for MeltanoLabs target-postgres

tayloramurphy:


_data/meltano/loaders/target-postgres/meltanolabs.yml
```diff
@@ -0,0 +1,14 @@
...
+description: PostgreSQL database loader
+namespace: target_postgres
+dialect: postgres
+target_schema: $TARGET_POSTGRES_SCHEMA
+logo_url: /assets/logos/loaders/postgres.png
+variant: meltanolabs
+repo: https://github.com/meltanolabs/target-postgres
+pip_url: git+https://github.com/meltanolabs/target-postgres.git
+domain_url: https://www.postgresql.org/
+maintenance_status: development
```
pnadolny13, line 12: @tayloramurphy theres no capabilities or settings, is that on purpose because it's still in development?

tayloramurphy: @pnadolny13 yes, that's on purpose. I expect we'll get Derek or someone else to update it. I'm just making the initial swap.

pnadolny13 (approved):
@tayloramurphy makes sense! Looks good to me

PR #751, opened 2022-08-26
transferwise mongodb use strings for booleans

pnadolny13:
Related to slack thread https://meltano.slack.com/archives/C01TCRBBJD7/p1661503116004259

The code looks for the string value `true` vs an actual boolean.


tayloramurphy:
Indeed this looks right 🤦 https://github.com/transferwise/pipelinewise-tap-mongodb/blob/master/tap_mongodb/__init__.py#L259

tayloramurphy (approved):

PR #752, opened 2022-08-26
add star button

pnadolny13:
I know we're migrating to gridsome soon but I found https://webapps.stackexchange.com/questions/79460/how-do-i-make-a-star-this-project-on-github-link and https://buttons.github.io/. I don't have too much FE experience but it looked like an easy add and my local build looked pretty decent so 🤷 .

Not sure if we wanted to point to the hub repo vs meltano.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -43,4 +43,6 @@
 
     gtag('config', 'G-NSVMGNSZKN');
   </script>
+  <!-- Place this tag in your head or just before your close body tag. -->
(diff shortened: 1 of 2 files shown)


tayloramurphy:
@pnadolny13 we're launching the gridsome revamp next week. I'm ok with adding this on classic for now but not guaranteed to carry over.

tayloramurphy (approved):

PR #768, opened 2022-09-11
Render inline markdown from plugin data

### LINKED ISSUE #765: feat(gridsome): support for markdown rendering from **`yaml`/`graphQL` page content**
Related to 

- #764

rwfeather:
Closes #765 

These changes will render the markdown found in the following plugin data sections:

- `settings.description` (common)
- `settings_preamble` (rare)
- `usage` (rare)
- `prereq` (rare)
- `next_steps` (rare)

For testing/review purposes, the superset utility data includes all of the above sections.

The styling of these sections may need some additional work, but this at least adds the rendered markdown.

### DIFF
--- a/gridsome.server.js
+++ b/gridsome.server.js
@@ -29,20 +29,20 @@ function renderMarkdownSections(pluginData) {
       pluginData.settings &&
       pluginData.settings.map((setting) => ({
         ...setting,
-        descriptionRendered:
+        description_rendered:
           setting.description && marked.marked(setting.description),
       })),
     // Rare
-    settings_preambleRendered: pluginData.settings_preamble
+    settings_preamble_rendered: pluginData.settings_preamble
       ? marked.marked(pluginData.settings_preamble)
       : undefined,
-    usageRendered: pluginData.usage
+    usage_rendered: pluginData.usage
       ? marked.marked(pluginData.usage)
       : undefined,
-    prereqRendered: pluginData.prereq
+    prereq_rendered: pluginData.prereq
       ? marked.marked(pluginData.prereq)
       : undefined,
-    next_stepsRendered: pluginData.next_steps
+    next_steps_rendered: pluginData.next_steps
       ? marked.marked(pluginData.next_steps)
       : undefined,
   };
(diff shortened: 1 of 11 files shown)


rwfeather:
Ready for review. Compare the `superset` utility to see changes:

[before](https://meltano-hub-gridsome.netlify.app/utilities/superset/apache/)

[after](https://deploy-preview-768--meltano-hub-gridsome.netlify.app/utilities/superset/apache/)

aaronsteers (approved):
Nice, @rwfeather. I learned a bunch from reading this PR. 

As you already called out, there are some changes we'll need to make on the surrounding text/context, but that can follow after.

Did you have any preference of `*Rendered` vs `_rendered` as suffix?

I also would not be against simply overriding the value or the raw markdown source data with the rendered html output.

All of the above is pretty easy to iterate on in the future, and nothing I can see worth blocking on. 

Approving so you are clear to merge to `gridsome-main` when ready.

rwfeather:
> Nice, @rwfeather. I learned a bunch from reading this PR.
> 
> As you already called out, there are some changes we'll need to make on the surrounding text/context, but that can follow after.
> 
> Did you have any preference of `*Rendered` vs `_rendered` as suffix?
> 
> I also would not be against simply overriding the value or the raw markdown source data with the rendered html output.
> 
> All of the above is pretty easy to iterate on in the future, and nothing I can see worth blocking on.
> 
> Approving so you are clear to merge to `gridsome-main` when ready.

Good point regarding the naming @aaronsteers , I've changed it to `_rendered` because that's much more consistent with existing graphql fields.

I thought about overwriting the existing fields, but was concerned that there might be a case in the future where a downstream component wanted access to the raw markdown. Just in case it wanted to render it differently somehow 🤷‍♀️ . There also is the possibility of using this graphql schema to drive things that aren't webpages (which was the original plan for `src/api`; to generate the JSON API from the graphql). I think I'll leave it as-is because this naming scheme makes it clear that the `rendered` fields contain raw html.

rwfeather:
Thanks for the quick review!

Issue #570, opened 2022-06-07
Rebuilt plugin-specific usage metrics


aaronsteers:
The prior usage metrics for plugins were not able to distinguish the actual variant being used, and therefor had incorrect (inflated) usage metrics. For instance, all `tap-postgres` implementations would have usage metrics of the full combined usage across each.

We'll want to rebuild these metrics now with variant-specific metrics, now that we will have access to this distinction.

pnadolny13:
This issue depends on https://github.com/meltano/squared/issues/275 in order to unhash variant names. 

Also I opened https://github.com/meltano/hub/issues/639 to get the current metrics updated to make it more clear that theyre not variant level.

tayloramurphy:
Thanks for opening the issue @pnadolny13. Moving this out a few iterations in that case.

pnadolny13:
Theres more detail on how we want to expose these numbers related to Success/Failure attribution in https://github.com/meltano/squared/issues/298#issuecomment-1187708092. I closed that one because it was a duplicate.

pnadolny13:
@aaronsteers @tayloramurphy I unassigned myself from this as I'm working on making the data available in https://github.com/meltano/squared/issues/298#issuecomment-1187708092.

Once its available, I believe @alexmarple or @rwfeather would be the one actually taking on this work to update how the hub uses it.

PR #769, opened 2022-09-12
Add metrics to plugin sidebars, start using variant-level metrics

### LINKED ISSUE #649: Revamp: Pull in Plugin statistics/metrics to detail pages
This iteration will render the data from `squared`'s S3 files into the Gridsome site.

Integration with or building of a API layer is out of scope.

rwfeather:
Closes #649 #570 

This adds metrics to the plugin sidebar, using the same badges as the old site (do we want to reformat them at all?)

Since it was fairly easy, I've also started pulling in the new `variant_metrics.yml` data provided by the data team.

It was unclear to me which of the data points we want to display on the new hub (`ALL_EXECS`, `LEGACY_ALL_EXECS`, `SUCCESS_EXECS`) 

```yml
https://github.com/anelendata/tap-exchangeratehost:
    SUCCESS_EXECS: 231
    LEGACY_ALL_EXECS: 303
    LEGACY_SUCCESS_PROJECTS: 10
    LEGACY_ALL_PROJECTS: 11
    SUCCESS_PROJECTS: 10
    PLUGIN_TYPE: extractors
    LEGACY_SUCCESS_EXECS: 251
    VARIANT: anelendata
    PIP_URL: git+https://github.com/anelendata/tap-exchangeratehost.git
    ALL_PROJECTS: 10
    ALL_EXECS: 283
    NAME: tap-exchangeratehost
```

For many plugins they had null values for 1, 2, or 3 of the options. Could @pnadolny13 be able to clear up the difference?


rwfeather:
Also, should I be concerned that `variant_metrics.yml` is smaller than `metrics.yml`?

Screen Shot 2022-09-11 at 8 35 55 PM

I honestly expected the opposite (all variants > just defaults)

aaronsteers:
@rwfeather - Since 'legacy' metrics are generally not actually accurate for the individual variant being displayed, I think the 'legacy' metrics should be ignored entirely.

Specifically, going forward, the 'number of downloads/installs/executions/projects' should not be displayed on a plugin page if those are global to all variants of the tap. That info might be interesting to us to gage overall interest in the upstream source, but it's not helpful to the user since.

aaronsteers:
@rwfeather - I can't say with confidence why the new variant-specific file would be larger, except perhaps that there may be a lot of variants and plugins overall that we don't yet have data from in post-2.0 telemetry.

@pnadolny13 - cc for confirm/review.

aaronsteers:
I checked Postgres real quick:

[The default pipelinewise variant page](https://deploy-preview-769--meltano-hub-gridsome.netlify.app/extractors/tap-postgres/transferwise/):

image

[The singer-io variant page](https://deploy-preview-769--meltano-hub-gridsome.netlify.app/extractors/tap-postgres/singer-io/):

image

These metrics look correct to me, and you can see easily from here that the pipelinewise variant has an order of magnitude more project references and a few order of magnitudes more executions.

@rwfeather - Do you think you can look into why the 'repo not found' error is printing for those badges on the pipelinewise side?

aaronsteers (approved):

rwfeather:
Sounds good 👍 I'll merge this because it seems to do what we want for now. Gonna see if I can fix the repo badges real quick. Thanks @aaronsteers!

pnadolny13:
>It was unclear to me which of the data points we want to display on the new hub (ALL_EXECS, LEGACY_ALL_EXECS, SUCCESS_EXECS)

@rwfeather I took the approach of keeping backwards compatibility in the new file by including both the legacy metrics (e.g. LEGACY_ALL_EXECS) and the new variant level metrics since I wasnt sure how/when you all would migrate to the new metrics. The LEGACY_ metrics use the old method of lumping all variants together. The difference between ALL_EXECS and SUCCESS_EXECS (which I think AJ might have already explained) is that success is a subset of all meaning some projects install and try to use the plugin but never have a successful execution so it doesnt get counted towards success. This way we can track the drop off between installing and successfully using a plugin.

>For many plugins they had null values for 1, 2, or 3 of the options.

Like AJ mentioned, lots of them either dont have usage data in post-2.0 yet or truly dont have usage at all. I found a bunch of the non-default variants that are in bad shape never get used at all.

>Also, should I be concerned that variant_metrics.yml is smaller than metrics.yml?

Nope not a concern, the difference is that previously we included metrics for all singer variants we found when scraping github whether they were listed on the hub or not. Now we only include metrics for variants that were found AND are listed on the hub. There was a lot more extra data in the old file since we werent pulling hub data into the warehouse when it was created so there was no way to filter.

Happy to add more context to any of this if needed!

PR #772, opened 2022-09-12
Make list pages only show defaults, change path sep

### LINKED ISSUE #767: Revamp: Fix url parameters and redirects
The default pages link to the URL with the variant in the path (i.e. https://meltano-hub-gridsome.netlify.app/extractors/tap-auth0/matatika/) when it should be (https://meltano-hub-gridsome.netlify.app/extractors/tap-auth0/).

Also the current site uses a `/plugin--variant` for the url as opposed to gridsome's `/plugin/variant`. We'll want to redirect the old links to the new links.

rwfeather:
Closes #767 

This PR:

- Makes it so that the plugin lists (`/extractors`, `/loaders` etc) display a single entry for each plugin (previously they would have a card for each available variant)
- Making the links on the list pages direct to the plugin path rather than the variant path (`/extractors/tap-csv` instead of `/extractors/tap-csv/transferwise`)
  - And also for search results dropdown
- Changes the path convention to match old hub (`/:plugintype/:name/:variant` -> `/:plugintype/:name--:variant`)

I prefer having the `/` separator between the name and variant because it matches our directory scheme a bit better. But it's probably wise to not break any existing links out in the wild.

### DIFF
--- a/gridsome.config.js
+++ b/gridsome.config.js
@@ -14,37 +14,37 @@ module.exports = {
   templates: {
     Extractors: [
       {
-        path: "/extractors/:name/:variant",
+        path: "/extractors/:name--:variant",
         component: "./src/templates/Extractors.vue",
       },
     ],
     Loaders: [
       {
-        path: "/loaders/:name/:variant",
+        path: "/loaders/:name--:variant",
         component: "./src/templates/Loaders.vue",
       },
     ],
     Orchestrators: [
       {
-        path: "/orchestrators/:name/:variant",
+        path: "/orchestrators/:name--:variant",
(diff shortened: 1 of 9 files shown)


aaronsteers:
> I prefer having the `/` separator between the name and variant because it matches our directory scheme a bit better. But it's probably wise to not break any existing links out in the wild.

Agreed on both points. The `/` is nicer and seems more idomatic for a web url, but agreed also prioritizing compatibility and not breaking existing links.

aaronsteers (commented):
I did a code review and browsed the preview site. All looks good 👍

aaronsteers (approved):

PR #781, opened 2022-09-13
refactor(gridsome): unified plugin data model and template page

### LINKED ISSUE #776: bug(gridsome): detail pages need `--variant` flag for install
Same as #774 but for the detail page content. (Whereas #774 was for the sidebar.)

As there in #774, I think we only want to use `--variant` if the plugin is not the default.

aaronsteers:
Closes #776 

Originally intended to update install instructions to add the `--variant` option, I realized that this was in 6 places for the 6 plugin types.

I decided to replace the multiple redundant detail page templats with a single detail page `Plugins.vue` and this required a refactor of the API to use `allPlugins()` as an entrypoint for all plugins.

The net result is that we have fewer pages to maintain and we can now query 'plugins' generically by type or by other fields.


schema.graphql
```diff
@@ -31,6 +31,42 @@ input BelongsToFilterInput {
 
 scalar Date
 
+type Plugins implements Node {
```
rwfeather, line 34: Not sure if you updated this by hand, but this was added by alex for easier reference (copypasted from the http://localhost:8080/___explore playground)

This file is not used at all in the build process, the schema is generated automatically. As far as i can tell there's no way for us to have a static schema file to be picked up by the build process.

rwfeather (approved):

schema.graphql
rwfeather: @aaronsteers I'd be fine with removing the file at this point. There's nothing that's in place to keep it updated anyway.

aaronsteers: Thanks, @rwfeather! I've removed the file per your comment.

PR #786, opened 2022-09-14
feat: Add tailwind to Hub using gridsome plugin

rwfeather:
This enables us to use tailwind utility classes in our components.

There is a downside however to adding tailwind to a project this far along. We might struggle to use tailwindcss' default (opinionated) mode: [Preflight](https://tailwindcss.com/docs/preflight)

To see what I mean, all you need to do is [browse the preview build](https://deploy-preview-786--meltano-hub-gridsome.netlify.app/) for this branch.

Preflight itself is pretty cool, and it would be a great base layer for us to build on if we managed to wrangle the current design to fit it. It plays really well with modern design systems, smooths over the rougher parts of the browser ecosystem, and would make it a lot easier to create a consistent look and feel across meltano's web front ends.

I see 2 approaches (or maybe two phases):

- [Disable Preflight](https://tailwindcss.com/docs/preflight#disabling-preflight), keeping our current setup intact while still letting us incrementally adopt utility classes where we need them. (#788 )
- Keep Prefilght and begin updating our existing styles and elements with utility classes (#787)

I've created two additional PR's that demonstrate each of the approaches so that anyone can check the Netlify previews.

### DIFF
--- a/src/components/PluginCapabilitiesSection.vue
+++ b/src/components/PluginCapabilitiesSection.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <p class="text-3xl" id="capabilities">Capabilities</p>
+    <p class="text-3xl py-4" id="capabilities">Capabilities</p>
     <span v-if="capabilities">
       <span
         >The current capabilities for
@@ -11,7 +11,7 @@
         capabilities for this {{ plugin_type }}.</span
       >
       <p>This plugin has the following capabilities:</p>
-      <ul>
+      <ul class="list-disc list-inside">
         <li v-for="(capability, index) in capabilities" v-bind:key="index">
           {{ capability }}
         </li>
--- a/src/components/PluginCapabilitiesSection.vue
+++ b/src/components/PluginCapabilitiesSection.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <h2 id="capabilities">Capabilities</h2>
+    <p class="text-3xl" id="capabilities">Capabilities</p>
     <span v-if="capabilities">
       <span
         >The current capabilities for
--- a/gridsome.config.js
+++ b/gridsome.config.js
@@ -7,7 +7,15 @@
 module.exports = {
   siteName: "Meltano Hub",
   icon: "./src/meltano-favicon-192x192.png",
-  plugins: [],
+  plugins: [
+    {
+      use: "gridsome-plugin-tailwindcss",
+
+      options: {
+        tailwindConfig: "./tailwind.config.js",
+      },
+    },
+  ],
   chainWebpack: (config) => {
     config.resolve.alias.set("@logos", "@/../assets/logos");
   },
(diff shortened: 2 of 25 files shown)


src/components/PluginCapabilitiesSection.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <h2 id="capabilities">Capabilities</h2>
+    <p class="text-3xl py-4" id="capabilities">Capabilities</p>
```
aaronsteers, line 3: Is it standard to not use any header labels or headset classes at all?

aaronsteers:
Troubleshooting block has an odd off-color background:

image

Similar off-color background on the settings descriptions:

image

aaronsteers:
Can we use a normal dark font in the search box? It seems to be defaulting to white-on-grey as of now:

image

src/components/PluginCapabilitiesSection.vue
pandemicsyn: with preflight headings are un-styled by default: https://tailwindcss.com/docs/preflight#headings-are-unstyled

tbh - I've only ever worked in configs where someone else had already added them as part of a more complex custom base style, and for my random side project's i've just always used the tailwind typography plugin. 

tl;dr this is fine I think until we have time to extend the base style?

pandemicsyn (approved):
I didn't do an exhaustive walkthrough of the deploy preview but the changes look go to me (with the exception of the call outs @aaronsteers made for the components where we used that slate background...although I kinda like on the settings actually lol)

Nice job @rwfeather, guessing it was a bit of slog tracking all this down  😁

src/components/PluginCapabilitiesSection.vue
cjohnhanson: I don't have a strong opinion here but if we wanted to keep this DRY and ensure consistency but don't think this merits an actual component, I think tailwind supports a middle path option within its plugin system where we could factor `text 3xl py-4` into a class: https://tailwindcss.com/docs/functions-and-directives#using-apply-with-per-component-css.

That said, I think the approach @rwfeather took here is the approach that tailwind recommends.

cjohnhanson:
Looks good! Like @pandemicsyn I also kind of like the slate background for the settings descriptions though maybe with a little extra padding on the left.

cjohnhanson (approved):

rwfeather:
The slate backgrounds were something I added to differentiate page content that comes from markdown blocks inside the yaml data. Most of the time they call out information that is really specific to the plugin you're looking at and I feel like they draw the eye and help break up the settings section a little bit visually.

I've padded them slightly so it looks a little better.

aaronsteers:
@rwfeather - Thanks so much. I felt less strongly about the setting blocks and the larger issue for me was the Troubleshooting block. But the explanation of the rendering this way because that is dynamic yaml/markdown based content makes a lot of sense.

We can continue tweaking as we go. 🚀

PR #787, opened 2022-09-14
Replace existing styles/elements with idiomatic Tailwind approaches

rwfeather:
If we want to get to the point where we can use Prefight (I argue that we should, eventually) it would start to look like this.

- Addressing each of the items in the Preflight doc (https://tailwindcss.com/docs/preflight) such as no longer using `<h1234>`, explicitly styling lists, specifying margins, etc.
- Yanking out the parts of `main.css` that conflict with the above, and I ran into quite a few of them while exploring this
- Using https://tailwindcss.com/docs/typography-plugin for styling rendered markdown sections

### DIFF
--- a/src/components/PluginCapabilitiesSection.vue
+++ b/src/components/PluginCapabilitiesSection.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <h2 id="capabilities">Capabilities</h2>
+    <p class="text-3xl" id="capabilities">Capabilities</p>
     <span v-if="capabilities">
       <span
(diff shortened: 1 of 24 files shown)


rwfeather:
It's possible I'm overestimating the amount of work this approach will take to get to parity. And there are definitely some parts of the site that _already_ look better with Preflight enabled.

Just didn't want to go too far down this road without a quick discussion, given our timeline.

rwfeather:
Now that we've decided to move forward with Preflight, I've consolidated back down to 1 PR

PR #788, opened 2022-09-14
Disable Tailwind Preflight

rwfeather:
This demonstrates what the site would look like if we add tailwind but can't invest the time into using their base layer, [Preflight](https://tailwindcss.com/docs/preflight#disabling-preflight).


aaronsteers:
Per our conversation in slack, looks like we're going in this direction: 

- #787 

We can still reopen/restore this PR if we decide to change course.

PR #791, opened 2022-09-15
change: remove singer-specific API in favor of common plugin API (BREAKS)

aaronsteers:


tayloramurphy:
@aaronsteers I think we need to chat about this before we pull the plug on this API. We'd committed last year to offering a trimmed down version of the API just for Singer, and it seems like that should be fairly simple to do.

tayloramurphy (changes requested):
Requesting changes as we've not made the call to drop the singer specific api endpoints.

tayloramurphy (approved):

PR #794, opened 2022-09-15
change(gridsome): remove tap and target indexes

aaronsteers:


aaronsteers:
We'll want to add the following redirects:

- `https://hub.meltano.com/singer/taps/` > `https://hub.meltano.com/extractors/`
- `https://hub.meltano.com/singer/targets/` > `https://hub.meltano.com/loaders/`

@rwfeather - Is there a simple means to add these redirects?

rwfeather:
> @rwfeather - Is there a simple means to add these redirects?

@aaronsteers Two methods I think:

- Update the netlify config `netlify.toml` with a [redirect rule](https://docs.netlify.com/routing/redirects/#syntax-for-the-netlify-configuration-file) using a [splat in the path](https://docs.netlify.com/routing/redirects/redirect-options/#splats) 
- Create "pages" that have the desired path (`/singer/taps/*`, `/singer/target/*`) while using the same underlying component to render the page. You can see an example of this method in `gridsome.server.js 187:219`

```js
  // Create default variant pages
  api.createPages(async ({ createPage, graphql }) => {
    const defaultPlugins = await graphql(`...`)

    Object.keys(defaultPlugins.data).forEach((query) => {
      // For each default plugin we create an extra page one-level up
      defaultPlugins.data[query].edges.forEach(({ node }) => {
        createPage({
          // Remove the variant part of the path
          path: node.path.split("--")[0],
          // And send it to the same rendering template
          component: `./src/templates/Plugins.vue`,
          context: {
            id: node.id,
            // This is the path given to the `page-query` to do the lookup
            path: node.path,
            name: node.name,
          },
        });
      });
    });
```

rwfeather:
For the second method, If there's any conditional rendering we want to do on the plugin page, it'll be possible to add something like an `isSinger` to the context object.

First method will be a lot lower-code and easier.

rwfeather (approved):

PR #798, opened 2022-09-15
feat: Hide search results when not in focus

### LINKED ISSUE #777: Revamp: Search bar doesn't respond to most page navigation, focus changes
- The search results dropdown doesn't disappear when clicking a search result, page navigation just happens in the background
- It should also disappear when the search input loses focus (someone clicks another part of the page)

The relevant component is in `src/components/Search.vue`

@cjohnhanson I wrote this component but didn't have much of a chance for follow-up enhancements. Hopefully this is something easy to pick up but if not please don't hesitate to throw it back to me.

cjohnhanson:
Turns out algolia was kind of overkill for this, the gridsome plugin only provides the search logic and doesn't actually provide a search bar component like I had hoped. 

These changes make it so that the search results are hidden whenever the search bar is not focused. It also unfocuses the search bar and clears the input after a result is clicked.

Closes #777


src/components/Search.vue
```diff
@@ -29,6 +40,7 @@
               />
               <h1>
                 {{ plugin.node.label }}
+                J
```
rwfeather, line 43: extra J

src/components/Search.vue
```diff
@@ -41,6 +53,7 @@
         <p>Your search didn't return any results. Please try again.</p>
       </div>
     </div>
+    o
```
rwfeather, line 56: extra o

rwfeather (approved):
lgtm 👍

aaronsteers (approved):
I tested successfully. Once the two notes from @rwfeather are resolved, this looks ready to merge.  👍

PR #808, opened 2022-09-19
Add Google Analytics to gridsome site

### LINKED ISSUE #805: Ensure GA parity in Hub Revamp
Before launching, we need to confirm that GA metrics are arriving as expected, at least parity with legacy Hub site.

cc @rwfeather, @pnadolny13, @tayloramurphy 

Related:

- https://github.com/meltano/internal-general/issues/444 (internal link)

rwfeather:
Closes #805 

This PR adds Google Analytics capabilities to the new gridsome based site. I've tested it out in the browser devtools, compared it to current hub behavior, and it seems to be working correctly.

@aaronsteers do you know who would be able to help confirm this is working on the GA dashboard?

### DIFF
--- a/gridsome.config.js
+++ b/gridsome.config.js
@@ -16,6 +16,14 @@ module.exports = {
       },
     },
     {
+      use: "gridsome-plugin-gtag",
+      options: {
+        config: {
+          id: "G-NSVMGNSZKN",
+        },
+      },
+    },
+    {
       use: "@gridsome/source-filesystem",
       options: {
         typeName: "MarkdownDocs",
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "@tailwindcss/typography": "^0.5.7",
     "graphql-tag": "^2.12.6",
     "gridsome": "^0.7.0",
+    "gridsome-plugin-gtag": "^0.1.10",
     "js-yaml": "^4.1.0",
     "marked": "^4.1.0",
     "tailwindcss": "^3.1.8",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4903,6 +4903,13 @@ graphql@^14.4.2:
   dependencies:
     iterall "^1.2.2"
 
+gridsome-plugin-gtag@^0.1.10:
+  version "0.1.10"
+  resolved "https://registry.yarnpkg.com/gridsome-plugin-gtag/-/gridsome-plugin-gtag-0.1.10.tgz#99ab1b494a874a996b08ee7073766c8132614d59"
+  integrity sha512-/T1snhRRO4akDx6Hi4rYpMtXo8jo8uEf0brHbOI6bbb+vg7XTFucEAOKklUdKFTLKFAwH1VIQ+AZajU88U6iHg==
+  dependencies:
+    vue-gtag "^1.10.0"
+
 gray-matter@^4.0.2:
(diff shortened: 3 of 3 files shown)


rwfeather:
This plugin uses [vue-gtag](https://matteo-gabriele.gitbook.io/vue-gtag/) under the hood.

If at some point we need to generate events inside components, it would look something like this:

```js
export default {
  name: 'MyComponent',

  methods: {
    login () {
      this.$gtag.event('login', { method: 'Google' })
    }
  }
}
```

aaronsteers:
@afolson, @pnadolny13 - Do one or both of you have visibility to confirm whether GA events are being correctly received for the new Hub?

pnadolny13:
@aaronsteers I have access to GA but I'm not exactly sure how to check if were getting events from this. The only thing I found was a hostname filter that has some of the other deploy preview URLs but I dont see `deploy-preview-808`. But I do see events correctly firing in my developer console so I'm not all that confident that I can make the judgement here.

Screen Shot 2022-09-20 at 10 23 16 AM

rwfeather:
@pnadolny13 interesting to see that we get events from the deploy previews. I was generating events from a localhost domain (which GA probably filters out at their collector). 

Getting the deploy preview to generate events with the 808 host is probably going to be enough to validate that this is working. In case it takes some more experimentation, is it possible for me to get access to our GA dashboard? If not a quick collab session would work too.

aaronsteers:
@pnadolny13:

> The only thing I found was a hostname filter that has some of the other deploy preview URLs but I dont see `deploy-preview-808`. 

Just want to make sure it is noted that the naming convention for gridsome hub previews is different vs legacy hub previews. Should be match this name pattern:

- [deploy-preview-808--meltano-hub-gridsome](https://deploy-preview-808--meltano-hub-gridsome.netlify.app/)

> But I do see events correctly firing in my developer console so I'm not all that confident that I can make the judgement here.

Since it sounds like you and Reina have both confirmed locally, should we go ahead and merge - perhaps while opening another issue to validate?

And/or - @afolson, @emilylucie - Are you able to confirm if we have gotten any GA events from `deploy-preview-808--meltano-hub-gridsome`?

rwfeather:
@aaronsteers yeah, I'd be fine with merging and doing a circle-back to check for events showing up in GA

aaronsteers (commented):
Approving for merge. Will open another issue to confirm that telemetry is coming through, and in the meanwhile we are seeing about getting @rwfeather access to GA so she can review directly.

aaronsteers (approved):

pnadolny13:
@rwfeather @aaronsteers following up here, I do see some events from this 808 hostname now. It was probably a slight delay from when I was sending events to when they were showing up and Reina was using localhost which sounds like it might be getting filtered.

Screen Shot 2022-09-22 at 9 39 44 AM

>@pnadolny13 interesting to see that we get events from the deploy previews. I was generating events from a localhost domain (which GA probably filters out at their collector).

@rwfeather I was using the preview when I saw it firing events properly. So I guess thats probably me showing up in GA if they have localhost filters.

>In case it takes some more experimentation, is it possible for me to get access to our GA dashboard? If not a quick collab session would work too.

I dont know how to give you access. Its connected to my meltano google account so an admin probably needs to get you added.

PR #809, opened 2022-09-20
feat: Revamped plugin index pages

### LINKED ISSUE #800: change(gridsome): Card view or similar for index page grids
The index page layout is too tight as of now.

1. Entries in the matrix should have some padding to space entries between each other.
2. The icon space for each should used a fixed hight buffered by white space as needed, along with a max width.
   - This should force a consistent starting `y` position for each item's label in a row.
4. The hover animation should not resize or relocate items beneath or around the selected item.
5. As page resizes, num cols in each row should be constant. (Resize from 5 items per row down to 4 items per row should affect _all_ rows at the same time.

@tayloramurphy - Can you attach the designer's mockup?

cjohnhanson:
This rewrites each of the plugin index pages to make them more easily readable. It also makes the plugin listings more responsive.

Closes #800

### DIFF
--- a/src/pages/Extractors.vue
+++ b/src/pages/Extractors.vue
@@ -8,30 +8,43 @@
         learn more about extracting and loading data using Meltano, refer to the Data Integration
         (EL) guide.
       </p>
-      <ul class="list-disc list-inside plugins-list">
-        <li v-for="edge in $page.allPlugins.edges" :key="edge.node.id" class="page-single-plugin">
-          <g-link
-            class="items-center flex flex-auto flex-col justify-around"
-            :to="edge.node.path.split('--')[0]"
-          >
-            <g-image
-              v-if="edge.node.logo_url"
-              :src="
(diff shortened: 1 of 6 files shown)


aaronsteers:
Wow - this looks SO nice!

image

aaronsteers:
There's a problem with the page links at bottom of page - rending white text on light grey background. Either moving the next page links into the blue background area, or darkening the text color should remedy.

image

aaronsteers (approved):
Approving ahead of the mob season tomorrow. Great work!!

tayloramurphy:
@cjohnhanson looks like a great iteration. My notes:

* the pagination issue AJ already mentioned
* The height of each of the grids is dependent on a row based on the largest icon on that row. I think it's most noticeable on the loaders page where the shift in the cards between rows is more obvious https://deploy-preview-809--meltano-hub-gridsome.netlify.app/loaders/ I think just using the same size across the board would be cleanest?
* Is there a setting to have an incomplete row be centered? https://deploy-preview-809--meltano-hub-gridsome.netlify.app/orchestrators feels weird

PR #819, opened 2022-09-21
feat: standardize plugin card height, improve pager readability, center index pages

cjohnhanson:
This resolves most of the feedback given [here](https://github.com/meltano/hub/pull/809#issuecomment-1252541212).

@tayloramurphy I wasn't able to find a quick and easy way to center incomplete rows. I'll spend a little bit more time trying some things out this morning. @rwfeather do you have any ideas here (see that linked PR comment for context).

### DIFF
--- a/src/pages/Extractors.vue
+++ b/src/pages/Extractors.vue
@@ -1,6 +1,6 @@
 <template>
   <Layout>
-    <div class="plugins-overview">
+    <div class="plugins-overview mx-auto">
       <p class="text-3xl py-4">Extractors</p>
       <p>
         Meltano lets you easily extract data out of arbitrary sources (databases, SaaS APIs, and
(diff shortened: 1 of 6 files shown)


rwfeather:
@cjohnhanson been playing with the grid options for a bit, and I'm not sure how to accomplish this across all of our grids. Gonna keep trying for a bit longer.

Looking at `/orchestrators` vs. `/loaders` I think the issue is less prominent when we have a bunch of items and _just_ the final row is less than 4. 

Since the most glaring issue is when we don't have enough to fill the top row (orchestrators, the final extractors page) I'm wondering if a viable workaround might be to conditionally render something different if our _total_ items are <4 rather than worrying about it in the more general case.

rwfeather:
It would feel dirty, but I'm starting to think it's an intentional flex vs grid thing, rather than a tailwind thing we're both missing.

Thoughts?

cjohnhanson:
> @cjohnhanson been playing with the grid options for a bit, and I'm not sure how to accomplish this across all of our grids. Gonna keep trying for a bit longer.
> 
> Looking at `/orchestrators` vs. `/loaders` I think the visual issue is less prominent when we have a bunch of items and _just_ the final row is less than 4.
> 
> Since the most glaring issue is when we don't have enough to fill the top row (orchestrators, the final extractors page) I'm wondering if a viable workaround might be to conditionally render something different if our _total_ items are <4 rather than worrying about it in the more general case.

That's a really good idea. @tayloramurphy what do you think? The result here would be that final rows with <4 cards in multi-row results pages would still start on the left and have empty spaces on the right but in pages like the "orchestrators" page where there's <4 items _total_ the items would be centered.

aaronsteers:
For my part, I think the rendering here looks great. The grid's canvas is properly sized and centered even with only a single item.

Before:

image


After:

image

PR #822, opened 2022-09-21
chore: add airflow extension hub entry

pandemicsyn:
Notable differences from the orchestrator hub entry:

```yaml
# rather than pointing directly airflow
repo: https://github.com/meltano/airflow-ext 
# explicity setting the default executable to the wrapper
executable: airflow_invoker
# using a meltano variant rather than apache
variant: meltano
# the predefined commands are two extension commands, since the wrapper handles all regular invocations
commands:
  describe:
    executable: airflow_extension
    args: describe
  initialize:
    executable: airflow_extension
    args: initialize
```

There's also two new extension-specific settings additions:

```yaml
- name: extension.airflow_home
  label: Airflow Home
  value: $MELTANO_PROJECT_ROOT/airflow
  env: AIRFLOW_HOME
  description: |
    The directory where Airflow will store its configuration, logs, and other files.
- name: extension.airflow_config
  label: Airflow Home
  value: $MELTANO_PROJECT_ROOT/airflow/airflow.cfg
  env: AIRFLOW_CONFIG
  description: |
    The path where the Airflow configuration file will be stored.
```

## Reviewer Checklist


pandemicsyn:
@aaronsteers @tayloramurphy @pnadolny13 here's a staged hub entry for the new airflow ext. I think we can probably just keep this draft/pending until we're ready for folks to start using it.  This is mostly the same def as the original except for the things I mentioned above in the PR body. 

Few things I wasn't sure of and so just guessed at:

1. Which repo to point Airflow vs the extension. 
2. Variant

_data/meltano/utilities/airflow/meltano.yml
```diff
@@ -0,0 +1,81 @@
...
+repo: https://github.com/meltano/airflow-ext
+pip_url: git+https://github.com/meltano/airflow-ext.git@main apache-airflow==2.3.3 --constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.3.3/constraints-no-providers-3.8.txt
+executable: airflow_invoker
+variant: meltano
+maintenance_status: active
+commands:
+  describe:
+    executable: airflow_extension
+    args: describe
+  initialize:
```
pandemicsyn, line 14: This might still change to plain `init` before merge (re: https://github.com/meltano/edk/discussions/23)

_data/meltano/utilities/airflow/meltano.yml
```diff
@@ -0,0 +1,81 @@
+name: airflow
+namespace: airflow
+label: Airflow
+docs: https://docs.meltano.com/guide/orchestration
+repo: https://github.com/meltano/airflow-ext
+pip_url: git+https://github.com/meltano/airflow-ext.git@main apache-airflow==2.3.3 --constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.3.3/constraints-no-providers-3.8.txt
+executable: airflow_invoker
+variant: meltano
+maintenance_status: active
+commands:
```
tayloramurphy, line 10: Can you add in the commands from https://hub.meltano.com/orchestrators/airflow#commands ?

_data/meltano/utilities/airflow/meltano.yml
```diff
@@ -0,0 +1,81 @@
+name: airflow
+namespace: airflow
+label: Airflow
+docs: https://docs.meltano.com/guide/orchestration
+repo: https://github.com/meltano/airflow-ext
```
tayloramurphy, line 5: Let's keep this as the apache repo. I don't want people to come to https://deploy-preview-822--meltano-hub.netlify.app/utilities/airflow and walk away thinking Meltano maintains its own fork of Airflow. It should be clear this is the full Airflow tool and we've got an extension that makes it easy to use with Meltano. 

I think it'd be worth it to add a new key for `ext_repo`. In a separate PR we (not you) can get the frontend updated to render an extension link and maybe even an EDK badge.

_data/meltano/utilities/airflow/meltano.yml
```diff
@@ -0,0 +1,81 @@
...
+  label: Airflow Home
+  value: $MELTANO_PROJECT_ROOT/airflow/airflow.cfg
+  env: AIRFLOW_CONFIG
+  description: |
+    The path where the Airflow configuration file will be stored.
+logo_url: /assets/logos/orchestrators/airflow.png
+settings_preamble: |
+  Meltano [centralizes the configuration](https://docs.meltano.com/guide/configuration) of all of the plugins in your project, including Airflow's. This means that if the [Airflow documentation](https://airflow.apache.org/docs/apache-airflow/stable/howto/set-config.html) tells you to put something in `airflow.cfg`, you can use `meltano config`, `meltano.yml`, or environment variables instead, and get the benefits of Meltano features like [environments](https://docs.meltano.com/concepts/environments).
+
+  Any setting you can add to `airflow.cfg` can be added to `meltano.yml`, manually or using `meltano config`. For example, `[core] executor = SequentialExecutor` becomes `meltano config airflow set core executor SequentialExecutor` on the CLI, or `core.executor: SequentialExecutor` in `meltano.yml`. Config sections indicated by `[section]` in `airflow.cfg` become nested dictionaries in `meltano.yml`.
```
tayloramurphy, line 81:
```suggestion
settings_preamble: |
  Meltano [centralizes the configuration](https://docs.meltano.com/guide/configuration) of all of the plugins in your project, including Airflow's. This means that if the [Airflow documentation](https://airflow.apache.org/docs/apache-airflow/stable/howto/set-config.html) tells you to put something in `airflow.cfg`, you can use `meltano config`, `meltano.yml`, or environment variables instead, and get the benefits of Meltano features like [environments](https://docs.meltano.com/concepts/environments).

  Any setting you can add to `airflow.cfg` can be added to `meltano.yml`, manually or using `meltano config`. For example, `[core] executor = SequentialExecutor` becomes `meltano config airflow set core executor SequentialExecutor` on the CLI, or `core.executor: SequentialExecutor` in `meltano.yml`. Config sections indicated by `[section]` in `airflow.cfg` become nested dictionaries in `meltano.yml`.
keywords:
  - edk
```

_data/meltano/utilities/airflow/meltano.yml
```diff
@@ -0,0 +1,81 @@
+name: airflow
+namespace: airflow
+label: Airflow
+docs: https://docs.meltano.com/guide/orchestration
+repo: https://github.com/meltano/airflow-ext
+pip_url: git+https://github.com/meltano/airflow-ext.git@main apache-airflow==2.3.3 --constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.3.3/constraints-no-providers-3.8.txt
+executable: airflow_invoker
+variant: meltano
+maintenance_status: active
```
tayloramurphy, line 9:
```suggestion
maintenance_status: active
definition: is an [orchestrator](https://docs.meltano.com/concepts/plugins#orchestrators) that allows for workflows to be programmatically authored, scheduled, and monitored.
```

_data/meltano/utilities/airflow/meltano.yml
pandemicsyn: 👍 yep let me bring them over. I assumed we didn't need those anymore but in hindsight makes sense that we'd keep them around.

_data/meltano/utilities/airflow/apache.yml
```diff
@@ -0,0 +1,91 @@
...
+pip_url: git+https://github.com/meltano/airflow-ext.git@main apache-airflow==2.3.3 --constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.3.3/constraints-no-providers-3.8.txt
+executable: airflow_invoker
+variant: apache
+maintenance_status: active
+definition: is an [orchestrator](https://docs.meltano.com/concepts/plugins#orchestrators) that allows for workflows to be programmatically authored, scheduled, and monitored.
+commands:
+  describe:
+    executable: airflow_extension
+    args: describe
+  initialize:
```
pnadolny13, line 16: @pandemicsyn @tayloramurphy the initialize command functionality is slightly confusing for me on first glance because it seems like its mixing both initializing the db and the meltano project assets. The project assets only need to initialized once then git versioned but the db stuff needs to be done in each environment, right?

Maybe all we need is an extra blurb that describes what the initialize command does and when you should run it (i.e. once per project or once per environment). That would take some of the mystery out of it.

pnadolny13 (changes requested):
@pandemicsyn this is looking good! Sorry it took so long for me to test it out. A couple pieces of feedback:

1. I was able to install and run `initialize` but when I ran the "user create" command I got an error:

```
(meltano) Patricks-MBP:data pnadolny$ meltano --log-level=debug invoke airflow users create -u admin@localhost -p password --role Admin -e admin@localhost -f admin -l admin
Traceback (most recent call last):
  File "/Users/pnadolny/Documents/Git/GitHub/meltano/squared/data/.meltano/utilities/airflow/venv/lib/python3.9/site-packages/meltano/edk/extension.py", line 119, in pass_through_invoker
    self.pre_invoke(None, command_args)
TypeError: pre_invoke() takes 1 positional argument but 3 were given
pre_invoke failed with uncaught exception, please report to maintainer
```

2. I see that the initialize command creates an `airflow` directory in my project root and also creates a `dags` folder in my `orchestrate` folder. Was there a specific reason for doing it that way? Did we decide to pull the config files out of `.meltano` so users can version control them? If so, I wonder if it would be better to put everything under the same orchestrate directory either with `airflow`/`dags` at the same level or honestly `dags` under `airflow` makes more sense to me, maybe the current contents of `airflow` could be under a subdir called `config` or similar?

pandemicsyn:
> I was able to install and run initialize but when I ran the "user create" command I got an error:

Can you try updating, looks like you've got the latest version of the extension - but maybe an older version of the EDK ?

> 2\. Did we decide to pull the config files out of `.meltano` so users can version control them?

Yep, that had been mentioned a few times and so ended up doing that for the original airflow ext PR and carried that forward for Superset as well. Its configurable though, so can point it anywhere. We drop a .gitignore https://github.com/meltano/airflow-ext/issues/9 to help folks out regardless of where that gets pointed.  

>  If so, I wonder if it would be better to put everything under the same orchestrate directory either with airflow/dags at the same level or honestly dags under airflow makes more sense to me, maybe the current contents of airflow could be under a subdir called config or similar?

I slighly prefer organizing under plugin even if it was under a config dir but any of these sound fine to me. Happy to swithc to whatever pattern (@aaronsteers any prefs?)

_data/meltano/utilities/airflow/apache.yml
pandemicsyn: I think this is way @aaronsteers mentioned adding discrete init and setup in a follow up version: https://github.com/meltano/edk/discussions/23

I can definitely clarify what's going to happen in the docs.  Could also add a dedicated `initdb`, `setup`, `etc`  command (be an easy change).

pandemicsyn: @pnadolny13 fyi forgot that airflow extension works the same as the non-extension version. The db and config get created "automatically" _everytime_ you invoke airflow ...if they don't already exist that is.

It's actually just the dag generator and gitignore that gets created when you call initialize (basically the files-airflow replacement).  If `initialize` is the first thing you call then all three get created at once. But if you go straight from installing airflow to adding a user - you won't deploy a dag generator. I'll update the docs to clarify.

pandemicsyn:
> > @pandemicsyn I tried deleting everything and re-adding but still ran into the same error.
> 
> Let me double check and make sure I actually updated everything correctly on my end, trying to keep all the PR's in sync has been a little iffy.

@pnadolny13 its my fault, did miss an update - have an update/fix over in the airflow extension repo: https://github.com/meltano/airflow-ext/pull/19

pnadolny13:
@pandemicsyn its working for me now! I was able to install and run everything locally as expected.

I created a PR in squared to start using that, although in production we use the default airflow helm chart that calls airflow directly so its really only used locally. The only thing I'd comment  on that I noticed in [my diffs in squared](https://github.com/meltano/squared/pull/443/files#diff-20f21fce4edc093f431000c0bd30e6b61d003de3a6d8cb97a9b25a4bbf880b07L9) was that we hard code python 3.8 in the constraints file name now vs previously templating it to use `MELTANO__PYTHON_VERSION`. If this was discussed somewhere already, great - ignore this, otherwise just wanted to call it out as a difference I noticed.

pandemicsyn:
> If this was discussed somewhere already, great - ignore this, otherwise just wanted to call it out as a difference I noticed.

Ah nope, thats totally an oversight and just a side effect of me trying to get everything playing nice earlier on. I'll send along an update 🙏

aaronsteers:
> > I was able to install and run initialize but when I ran the "user create" command I got an error:
> 
> Can you try updating, looks like you've got the latest version of the extension - but maybe an older version of the EDK ?
> 
> > 2. Did we decide to pull the config files out of `.meltano` so users can version control them?
> 
> Yep, that had been mentioned a few times and so ended up doing that for the original airflow ext PR and carried that forward for Superset as well. Its configurable though, so can point it anywhere. We drop a .gitignore [meltano/airflow-ext#9](https://github.com/meltano/airflow-ext/issues/9) to help folks out regardless of where that gets pointed.
> 
> > If so, I wonder if it would be better to put everything under the same orchestrate directory either with airflow/dags at the same level or honestly dags under airflow makes more sense to me, maybe the current contents of airflow could be under a subdir called config or similar?
> 
> I slighly prefer organizing under plugin even if it was under a config dir but any of these sound fine to me. Happy to swithc to whatever pattern (@aaronsteers any prefs?)

I'm late to this thread, but I agree with keeping config as repo artifacts (which users can ignore or update as needed). And I wanted to suggest we use `/orchestrate/airflow` as the home directory - basically just keeping things simply and having all plugin artifacts self-contained there in the same root folder path. 

Just a small tweak, but I guess another option we could go in would be would be to explicitly call out of subdirectory for the `AIRFLOW_HOM` directory containing config artifacts... either `/orchestrate/airflow/config/...` or `/orchestrate/airflow/home/...`. This keeps things all self-contained, while perhaps minimizing the number of files a user would see when browsing `/orchestrate/airflow`.

pandemicsyn:
> Just a small tweak, but I guess I also like the option mentioned above where we use a dedicated subdirectory for the `AIRFLOW_HOME` dir containing config artifacts... maybe `/orchestrate/airflow/config/...` or `/orchestrate/airflow/home/...`. This keeps things all self-contained, while perhaps minimizing the number of files a user would see when browsing `/orchestrate/airflow/`. We could also then put our own readme in that subfolder, explaining how those configs relate to meltano configs (meltano configs override them) and that they can ignore the directory until/unless they need special customizations.

@aaronsteers 👍 easy-peasy. Will send a long an update.

pandemicsyn:
> 

@aaronsteers just to clarify , where do you want the existing `dags` and `plugins` folders. Historically, those have always been `$MELTANO_PROJECT_ROOT/orchestrate/dags` and `$MELTANO_PROJECT_ROOT/orchestrate/plugins.` and I kept that the same for the extension.  Would you like them under the airflow directory as well. Basically this directory tree:

```
├── meltano.yml
├── orchestrate
│   └── airflow
│       ├── .gitignore
│       ├── airflow.cfg
│       ├── airflow.db
│       ├── dags
│       │   ├── README.md
│       │   └── meltano_dag_generator.py
│       ├── logs
│       │   ├── dag_id=meltano_yetanothertest_tap-gitlab-to-target-postgres-processed
│       │   │   └── run_id=scheduled__2022-10-11T00:00:00+00:00
│       │   │       └── task_id=meltano_yetanothertest_tap-gitlab-to-target-postgres-processed_task0
│       │   │           └── attempt=1.log
│       │   ├── dag_processor_manager
│       │   │   └── dag_processor_manager.log
│       │   └── scheduler
│       │       ├── 2022-10-12
│       │       │   └── meltano_dag_generator.py.log
│       │       └── latest -> /Users/syn/projects/meltano-projects/extdev/orchestrate/airflow/logs/scheduler/2022-10-12
│       ├── plugins
│       └── webserver_config.py
```

If so, do you want me to update the deployed .gitignore to allow checking in dags and plugins ?

aaronsteers:
@pnadolny13 - What do you think of the below?:

```
├── meltano.yml
├── orchestrate
│   └── airflow
│       ├── .gitignore
│       ├── dags
│           └── ...
│       ├── logs
│           └── ...
│       ├── plugins
│           └── ...
│       ├── config
│           └── README.md
│           └── airflow.db
│           └── airflow.cfg
│           └── webserver_config.py
```

aaronsteers:
> ...where do you want the existing `dags` and `plugins` folders. Historically, those have always been `$MELTANO_PROJECT_ROOT/orchestrate/dags` and `$MELTANO_PROJECT_ROOT/orchestrate/plugins` and I kept that the same for the extension. Would you like them under the airflow directory as well?

I didn't realize we weren't already putting them in an `airflow` subfolder (sorry, that's on me), but considering that new orchestrators are coming out soon (Meltano's own, cron, and dagster, for instance), I do think this is a pretty good default base directory pattern: `$MELTANO_PROJECT_ROOT/<PLUGIN_TYPE>/<PLUGIN_NAME>/`

Otherwise, I'm not sure we could keep `dagster` and `cron` resources properly insulated from each other.

pandemicsyn:
@aaronsteers yea , i actually think throwing plugins/dags under `orchestrate/airflow` makes sense. Just to wrap things up - can you confirm whether you want folks to be able to check `plugins` and `dags` into git by default? The .gitignore we deploy right now uses a "global deny all by default" rule and then only allows specific items:

https://github.com/meltano/airflow-ext/blob/main/files_airflow_ext/dot-gitignore

aaronsteers:
> @aaronsteers yea , i actually think throwing plugins/dags under `orchestrate/airflow` makes sense. Just to wrap things up - can you confirm whether you want folks to be able to check `plugins` and `dags` into git by default? The .gitignore we deploy right now uses a "global deny all by default" rule and then only allows specific items:
> 
> [meltano/airflow-ext@`main`/files_airflow_ext/dot-gitignore](https://github.com/meltano/airflow-ext/blob/main/files_airflow_ext/dot-gitignore?rgh-link-date=2022-10-12T23%3A32%3A45Z)

Thanks for explaining that, @pandemicsyn. I think I misunderstood before.

I think "yes" to allowing (and encouraging) `plugins` and `dags` to be committed to git. We'd still want to ignore logs.

pandemicsyn:
@aaronsteers one small gotcha with writing configs to a `config` dir. It's a little contra the airflow docs, which all reference files from within AIRFLOW_HOME which probably isn't a huge deal. 

However, unless I'm overlooking it in https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html you can't control where webserver_config.py gets written via env vars. So that would still end up in AIRFLOW_HOME.

aaronsteers:
@pandemicsyn - No worries. And I think that makes sense.

_data/meltano/utilities/airflow/apache.yml
```diff
@@ -72,7 +72,7 @@ settings:
   env: AIRFLOW__WEBSERVER__WEB_SERVER_PORT
 - name: extension.airflow_home
   label: Airflow Home
-  value: $MELTANO_PROJECT_ROOT/orchestrate/airflow/config
+  value: $MELTANO_PROJECT_ROOT/orchestrate/airflow
```
aaronsteers, line 75: 👍

aaronsteers (approved):

pandemicsyn:
> @pandemicsyn - No worries. And I think that makes sense.

@aaronsteers cool in that case, i think the hub side should be good to go with my latest changeset. I have https://github.com/meltano/airflow-ext/pull/21 over on the extension side that we'll need in order to make this all play nice together. It updates the deployed gitignore and ensures the new config path exists.

pandemicsyn:
@tayloramurphy @pnadolny13 @aaronsteers fyi - retested with a fresh install from the deploy preview and as far as i can tell everything worked. Here's the resulting fs tree:

```
├── orchestrate
│   └── airflow
│       ├── .gitignore
│       ├── config
│       │   ├── airflow.cfg
│       │   └── airflow.db
│       ├── dags
│       │   ├── README.md
│       │   └── meltano_dag_generator.py
│       ├── logs
│       │   ├── dag_id=meltano_yetanothertest_tap-gitlab-to-target-postgres-processed
│       │   │   └── run_id=scheduled__2022-10-13T00:00:00+00:00
│       │   │       └── task_id=meltano_yetanothertest_tap-gitlab-to-target-postgres-processed_task0
│       │   │           ├── attempt=1.log
│       │   │           └── attempt=2.log
│       │   ├── dag_processor_manager
│       │   │   └── dag_processor_manager.log
│       │   └── scheduler
│       │       ├── 2022-10-13
│       │       ├── 2022-10-14
│       │       │   └── meltano_dag_generator.py.log
│       │       └── latest -> /Users/syn/projects/meltano-projects/extdev/orchestrate/airflow/logs/scheduler/2022-10-14
│       └── webserver_config.py
```

I'm gonna go ahead flip this out of draft state.

pnadolny13:
@pandemicsyn I gave it another test and I think the only thing I noticed was that the gitignore was acting a little funky. Installing Airflow for the first time into a project only gives me a diff of the following:

Screen Shot 2022-10-14 at 3 42 26 PM

That excludes that dag generator that was dropped in the project and `airflow.cfg` still 🤔 .

Another call out - I know theres some complexity going on with the whole AIRFLOW_HOME situation that I dont know enough about but isnt this breaking the pattern that we usually have of putting run logs/local dbs/ etc. in the .meltano folder. I assumed we'd want to put anything that is "blast away-able" into that directory like how dbt writes its compile target in there instead of in the `transform/` directory. I'd defer to you all but from the outside I thought that was the pattern we encouraged.

_data/meltano/utilities/airflow/apache.yml
tayloramurphy: I updated the description on the commands - @pandemicsyn will you review?

pandemicsyn:
> Another call out - I know theres some complexity going on with the whole AIRFLOW_HOME situation that I dont know enough about but isnt this breaking the pattern that we usually have of putting run logs/local dbs/ etc. in the .meltano folder. I assumed we'd want to put anything that is "blast away-able" into that directory like how dbt writes its compile target in there instead of in the transform/ directory. I'd defer to you all but from the outside I thought that was the pattern we encouraged.

The db can go wherever - that's already a config option, so I can definitely switch that back to hide it in `.meltano`. Looks like we could hide the log dir too by [setting]( https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html?highlight=base_log#config-logging-base-log-folder) `logging.base_log_folder`. 

/cc @aaronsteers @tayloramurphy any pref's/feedback related to @pnadolny13 concern.

tayloramurphy:
> The db can go wherever - that's already a config option, so I can definitely switch that back to hide it in .meltano. Looks like we could hide the log dir too by [setting](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html?highlight=base_log#config-logging-base-log-folder) logging.base_log_folder.

@pandemicsyn I'd say go ahead and move the airflow.db and the logs to `.meltano`. I'd make the argument that if we don't want users to version control it then we shouldn't drop it in the main folder for them.

tayloramurphy:
@pnadolny13 how do you feel about our metrics for these EDK-based plugins? shouldn't be too bad to distinguish them, right?

aaronsteers:
I agree with the above - if it's easy enough to accomplish without messing up anything else, moving the default logs directory and also the airflow db into .meltano would be a good change.

pandemicsyn:
@aaronsteers @tayloramurphy @pnadolny13 

I segmented the dirs furthers. I ended up having to add another pair of config opts (`scheduler.child_process_log_directory`, `logging.dag_processor_manager_log_location`) in addition to `logging.base_log_folder`  because it turns out `logging.base_log_folder` isn't used by all logging options 🙄 . I did also confirm that webserver_config.py is hard-coded - https://github.com/apache/airflow/issues/10206 and so we're SOL there. It's always going to get dropped in AIRFLOW_HOME.

Here's the resulting disk layout:

```shell
(ext-dev) ➜  extdev git:(main) ✗ tree -I venv .meltano/utilities/airflow
.meltano/utilities/airflow
├── airflow.db
└── logs
    ├── dag_processor_manager
    │   └── dag_processor_manager.log
    └── scheduler
        ├── 2022-10-17
        │   └── meltano_dag_generator.py.log
        └── latest -> /Users/syn/projects/meltano-projects/extdev/.meltano/utilities/airflow/logs/scheduler/2022-10-17

5 directories, 3 files
(ext-dev) ➜  extdev git:(main) ✗ tree orchestrate
orchestrate
└── airflow
    ├── config
    │   └── airflow.cfg
    ├── dags
    │   ├── README.md
    │   └── meltano_dag_generator.py
    └── webserver_config.py
```

With this new layout we'll no longer need a gitignore 🥳. Once this new disk layout has been confirmed to look good to everyone, I'll send along a PR to remove the `.gitignore` deploy in github.com/meltano/ext-airflow. If you test this now, you'll still get an erroneous gitignore in AIRFLOW_HOME.

tayloramurphy (approved):

pnadolny13 (approved):
@pandemicsyn this looks good to me! Thanks for addressing all the comments/feedback

PR #829, opened 2022-09-22
fix: resolve missing logo_url entries for dbt transformers

aaronsteers:


aaronsteers:
@rwfeather - I scanned all of the index pages and these are the only ones missing `logo_url` entries. Rather than modify the GraphQL as originally proposed, this is more expedient fix.

In future, we should make `logo_url` required for new entries if it is not already.

aaronsteers:
From the preview build: 
image

rwfeather (approved):

PR #834, opened 2022-09-22
Standardize logo, text, alignment for plugin grids

### LINKED ISSUE #833: gridsome: for index page grid items, add strict `y` boundary between text and image
Following from feedback here: https://meltano.slack.com/archives/C043R4WV6TS/p1663869223504869

I think we should probably have a fixed 'y' boundary between image and text. This will keep text in a row all starting from the same y position. Also would allow us to ensure that the text label always fits (up to two lines of it anyway) in the grid space below the image.

Presumably, this may shrink some images which currently are using proportionally more height of the overall box.

We can also reduce the font size slightly if that helps reduce space pressure.

For contrast, here are two renderings with where logos have very different 'y' dimensionality:

image

cjohnhanson:
Closes #833

### DIFF
--- a/src/pages/Extractors.vue
+++ b/src/pages/Extractors.vue
@@ -67,7 +67,7 @@ query ($page: Int) {
   allPlugins(
     perPage: 100
     page: $page
-    sortBy: "label"
+    sortBy: "label_lowercase"
     order: ASC
     filter: { isDefault: { eq: true }, pluginType: { eq: "extractor" } }
   ) @paginate {
--- a/src/pages/Extractors.vue
+++ b/src/pages/Extractors.vue
@@ -15,13 +15,13 @@
         <div
           v-for="edge in $page.allPlugins.edges"
           :key="edge.node.id"
-          class="rounded-md shadow-lg text-slate-800 h-36 hover:bg-slate-100 p-2 md:p-4 bg-white overflow-hidden"
+          class="rounded-md shadow-lg text-slate-800 h-48 hover:bg-slate-100 p-2 md:p-4 bg-white overflow-hidden"
         >
           <g-link
-            class="grid grid-rows-6 align-self-center place-items-center h-full w-full"
+            class="grid grid-rows-6 align-self-center place-items-center h-40 w-full"
             :to="edge.node.path.split('--')[0]"
           >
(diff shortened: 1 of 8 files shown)


rwfeather:
Plugin grid seems tricky to get right, now the multi-line names are overlapping logos:

Screen Shot 2022-09-22 at 3 20 04 PM

Feels like something has got to give with the varying logo and text sizes.

src/pages/Extractors.vue
```diff
@@ -30,13 +30,11 @@
...
-                class="place-self-center object-scale-down"
+                class="place-self-center"
               />
             </div>
-            <div class="row-span-1 self-end object-scale-down">
-              <p
-                class="font-bold underline text-xs md:text-lg justify-self-center place-self-end object-scale-down text-clip"
-              >
+            <div class="grid content-end row-span-5 self-stretch h-full">
+              <p class="font-bold underline text-xs lg:text-md justify-self-center text-clip" e>
```
rwfeather, line 37: Extra text "e" in these

rwfeather (approved):

cjohnhanson:
> Plugin grid seems tricky to get right, now the multi-line names are overlapping logos:

@rwfeather yeah I think what we want here would be "make all cards the height of the the tallest content and put them in a 4-column grid". Sizing would be easier to do with flexbox but the grid obviously is easier to do with grid.

tayloramurphy (changes requested):
@cjohnhanson anything we can do about that overlap? I'd be cool with assuming 2 lines of text and keeping logos to the max size of CSV (or whatever the largest one is).

Also, I just noticed that the alphabetization is case sensitive Screen Shot 2022-09-22 at 4 30 40 PM

aaronsteers:
> I'd be cool with assuming 2 lines of text

Agreed. Let's assume 1 or 2 lines of text max.

cjohnhanson:
@tayloramurphy @aaronsteers I took another pass at this this morning with hard-coded heights and it's looking better at the screen sizes I checked out. I also updated sorting to be case insensitive.

tayloramurphy (approved):
Love the changes! Looks great

aaronsteers (approved):
Looks really good! Thanks, @cjohnhanson !

PR #835, opened 2022-09-22
Add variants to safelist workaround

rwfeather:
The workaround for hot-reloading tailwind classes hasn't been respecting variants like `md:`, `sm:`, `hover:`.

Adding this option will generate those when we are in safelist mode


aaronsteers (approved):

cjohnhanson (approved):
🙌

PR #854, opened 2022-09-29
For settings blocks, add the environment variable information and default values

rwfeather:
Adding more information to the settings section

Screen Shot 2022-09-28 at 7 33 06 PM

Screen Shot 2022-09-28 at 7 36 41 PM

For the default values, because they can be anything (string, int, array, object) in the yaml, the graphql based build system had a hard time handling them in a consistent manner. In order for this to work, I've just stringified all default values for display purposes.


tayloramurphy (approved):
@rwfeather this looks great!

PR #861, opened 2022-09-30
fix: Don't include transforms in search results

### LINKED ISSUE #860: Plugins with one or more 'hidden: true' variant appear to be missing from search
The `hidden: true` is used to flag variants that are deprecated and are not promoted. When we removed these variants, it appears the corresponding entry was removed entirely from search results.

There may be some other nuance involved here, but I was able to confirm with these examples:

- `tap-adwords` (`meltano` variant hidden, but others should show)
- `tap-salesforce` (`meltano` variant hidden, but others should show)

Note: I've not confirmed root cause, only what appears to be a pattern and possible explanation.

Initially reported here:

- https://github.com/meltano/hub/discussions/818#discussioncomment-3767458

cjohnhanson:
Turns out that any tap which had an associated transform was trying to display the transform plugin within search results, which meant it was doing a `require(`!!assets-loader?width=75!@logos/transforms/${plugin.node.name}.png)` and since we don't have logos for transforms, that require was failing and preventing any search results for the tap being displayed.

@aaronsteers -- should we be including transforms within search results in the hub? If so, I'll update this to add logos for each.

Closes #860

### DIFF
--- a/src/components/Search.vue
+++ b/src/components/Search.vue
@@ -122,7 +122,9 @@ export default {
 
 <static-query lang="graphql">
 query {
-  allPlugins(filter: { isDefault: { eq: true }, hidden: { ne: true } }) {
+  allPlugins(
+    filter: { isDefault: { eq: true }, hidden: { ne: true }, pluginType: { ne: "transform" } }
+  ) {
     edges {
       node {
         id


aaronsteers:
@cjohnhanson - No, let's exclude transforms from all UI. Those are on their way to be deprecated/removed.

cjohnhanson:
> @cjohnhanson - No, let's exclude transforms from all UI. Those are on their way to be deprecated/removed.

@aaronsteers cool, this can go out as-is then.

aaronsteers (approved):
Looks great! Thanks for the same day turnaround. 🚀

PR #865, opened 2022-10-02
add metabase plugin

pnadolny13:
@tayloramurphy I was able to get Metabase working using the container spec feature. Its still limited by https://github.com/meltano/meltano/issues/3405 so I had to add some env vars that are accidentally overridden from my personal setting that get passed through.

Its a bit weird having an empty pip url. It makes me wonder how the EDK will work in relation to container based plugins.


tayloramurphy:
@pnadolny13 to your point about the empty pip_url - that's partially because of a JSON validation, right? It seems like if we implemented https://github.com/meltano/meltano/discussions/6459 we could make it optional.

pnadolny13:
> @pnadolny13 to your point about the empty pip_url - that's partially because of a JSON validation, right? It seems like if we implemented [meltano/meltano#6459](https://github.com/meltano/meltano/discussions/6459) we could make it optional.

Yeah right now the json schemas require that pip_url is present. So the tests were failing but an empty string looks to pass, I wonder if theres other weird implications if the `meltano add` logic assumes a pip install which would fail for this.

pnadolny13:
@rwfeather @edgarrmondragon @cjohnhanson @aaronsteers I need a little help testing that this will work when I merge. I include an empty pip_url since this is container only. Will that causes installation failures? I tried to install by pointing my meltano instance at a local hub api instance (and the deploy preview) but neither worked because it looks like its requesting the correct index url, then following the links returned but the links returned always have the default https://hub.meltano.com/ base url so they fail since it doesnt exist there yet.

Any thoughts on how to test it? I imagine we'll probably want to have the hub api logic to propagate the non-default url throughout when its generating the static API. Potentially also evaluating a container only option where pip url isnt required.

aaronsteers:
@pnadolny13 - You should be able to leave pip_url out altogether. Can you confirm if that approach works? 

Regarding the API URL root, let's see if we can target a branch as root using raw URL syntax....

@tayloramurphy We may also want consider logging and/or prioritizing the `meltano add --from_ref=...` approach where we point to the URL for a lock file or plugin definition directly. That would make installation of plugins much easier, directly from their own repo or from a test definition.

pnadolny13:
> @pnadolny13 - You should be able to leave pip_url out altogether. Can you confirm if that approach works?

@aaronsteers the plugin does works in meltano without the pip_url defined if I install it myself but I'm wondering if the install will fail if the pip_url in this hub definition is null i.e. does meltano run `pip install {plugin_definition_json.get('pip_url')}` when something is added? In that case I would expect the add/installation of this plugin to fail. Right?

> Regarding the API URL root, let's see if we can target a branch as root using raw URL syntax....

I don't think I follow this part, can you explain more?

To add detail in case my first comment wasn't super clear, right now if I request the index from the preview app it includes metabase (great!) but the entry looks like:

https://deploy-preview-865--meltano-hub.netlify.app/meltano/api/v1/plugins/utilities/index

```
    "metabase": {
        "variants": {
            "metabase": {
                "ref": "https://hub.meltano.com/meltano/api/v1/plugins/utilities/metabase--metabase"
            }
        },
        "default_variant": "metabase",
        "logo_url": "https://hub.meltano.com/assets/logos/utilities/metabase.png"
    }
```

So when meltano follows the `ref` to get the plugin definition json, it fails because the prod url doesnt exist yet.

Does that make sense?

rwfeather:
@edgarrmondragon @pnadolny13 I've readied a fix for the above issue that's preventing you from testing this change

- #876 

Running `gridsome develop` still seems to skip generating the API for some reason, but I can confirm that this should work for netlify preview builds

edgarrmondragon (changes requested):
@pnadolny13 you might need to move the logo from `assets/logos/utilities/metabase.png` to `static/assets/logos/utilities/metabase.png`.

aaronsteers:
> does meltano run `pip install {plugin_definition_json.get('pip_url')}` when something is added?

If `pip_url` is simply not there, it should skip attempting to install anything via pip.

aaronsteers:
> I've readied a fix for the above issue that's preventing you from testing this change

Thanks, @rwfeather !

aaronsteers:
Looking good! https://deploy-preview-865--meltano-hub.netlify.app/meltano/api/v1/plugins/utilities/index


```js
{
// ...
   "metabase":{
      "variants":{
         "metabase":{
            "ref":"https://deploy-preview-865--meltano-hub.netlify.app/meltano/api/v1/plugins/utilities/metabase--metabase"
         }
      },
      "default_variant":"metabase",
      "logo_url":"https://deploy-preview-865--meltano-hub.netlify.app/assets/logos/utilities/metabase.png"
   },
// ...
}
```

edgarrmondragon (approved):

cjohnhanson (approved):

pnadolny13:
@rwfeather thanks! It looks like I'm able to test and everything now.

>If pip_url is simply not there, it should skip attempting to install anything via pip.

@aaronsteers I was able to test it and it looks like it does fail to install because this definition has an empty string for the pip_url, although the plugin definition and meltano.yml entry still get added. So I'm able to then spin it up but I still got an ugly install error.

Screen Shot 2022-10-05 at 4 30 51 PM

We have 2 conflicting rules right now:
- the hub plugin definition JSON schemas require `pip_url` to not be null
- meltano looks like it needs `pip_url` to be null in order to know to skip the "pip install"

_data/meltano/utilities/metabase/metabase.yml
```diff
@@ -0,0 +1,40 @@
+name: metabase
+namespace: metabase
+label: Metabase
+description: Metabase is the easy, open-source way for everyone in your company to ask questions and learn from data.
+docs: https://www.metabase.com/
+repo: https://github.com/metabase/metabase
+maintenance_status: active
+variant: metabase
+pip_url: ''
```
aaronsteers, line 9:
```suggestion
```

Distinction of leaving out entirely vs leaving blank.

@pnadolny13 - Can you try this way with the field left out?

aaronsteers:
My expectation is that `pip_url` should be optional everywhere, but when it is specified, it needs to be a valid completion to the command `pip install ...`.

pnadolny13:
> My expectation is that `pip_url` should be optional everywhere, but when it is specified, it needs to be a valid completion to the command `pip install ...`.

@aaronsteers ok that makes sense - I meant that the hub schema definition for plugins require it in https://github.com/meltano/hub/blob/ac95d95ca2bd94eea24b8ffd36ca0e18b75ba143/schemas/common/common.schema.json#L11 so CI tests fail. If thats not a requirements we can remove that from the schema though, I didnt know that.

aaronsteers:
> > My expectation is that `pip_url` should be optional everywhere, but when it is specified, it needs to be a valid completion to the command `pip install ...`.
> 
> @aaronsteers ok that makes sense - I meant that the hub schema definition for plugins require it in https://github.com/meltano/hub/blob/ac95d95ca2bd94eea24b8ffd36ca0e18b75ba143/schemas/common/common.schema.json#L11 so CI tests fail. If thats not a requirements we can remove that from the schema though, I didnt know that.

Cool. Yeah, totally makes sense. We can update that json schema as needed. 👍

I believe this requirement was removed back when we introduced docker support in commands. But it's not widely used at this point so I think this might be the first thing added to the hub without a pip_url.

edgarrmondragon:
> What do we think the best path is for the local metabase data like the db it creates? I put it in `$MELTANO_PROJECT_ROOT/analyze/metabase/` but its not something that would be checked into git so maybe its better to go in `$MELTANO_PROJECT_ROOT/.meltano/utilities/metabase/data/`? Or `$MELTANO_PROJECT_ROOT/.meltano/run/metabase/data`? Is there conventions for where things go in the `.meltano` folder?

@pnadolny13 You're right, `.meltano` is a better place for it. That's what superset does: https://github.com/meltano/hub/pull/865#issuecomment-1270286465

aaronsteers:
@pnadolny13 - The `.meltano` directory can definitely work, and that's also where we put hidden dbt artifacts by default. I assume we don't have a repo scaffold or extension as of yet, correct?

The extension (if we add one later) can install a custom `.gitignore` as well as other files.

schemas/common/common.schema.json
```diff
@@ -8,7 +8,6 @@
     "required": [
         "name",
         "namespace",
-        "pip_url",
```
aaronsteers, line 11: 👍

aaronsteers (approved):

aaronsteers:
My review doesn't cover everything, but for my part, this looks good! 👍 ✅

pnadolny13:
Thanks @aaronsteers @edgarrmondragon. I moved it into the `.meltano` dir. We can have a follow on to add functionality with the EDK, sounds like from https://meltano.slack.com/archives/C013Z450LCD/p1654800904252579 that @z3z1ma has already made progress on that.

PR #867, opened 2022-10-03
Add default codeowners to repository

rwfeather:
Discussion with @aaronsteers, we'll have myself and @cjohnhanson as codeowners for the repository.

Do we want to set it up differently for PRs that touch `_data`?


aaronsteers (approved):
Probably the review process for `_data` should be different, but I don't know exactly who is best to take point on that. 

That said, we can merge this as the first iteration and iterate from here.

PR #874, opened 2022-10-05
add tap-xero definition content

pnadolny13:
There wasnt much content for this plugin and the repo's readme is hard to figure out whats required. I quickly pulled out the important bits.


sbalnojan (approved):
@pnadolny13 looks good to me. Let's get this shipped, love the response time, I think I just saw a slack message about this an hour ago.

PR #877, opened 2022-10-05
Fix URL handling in plugin API generation

rwfeather:
Fix for https://github.com/meltano/hub/pull/876#issuecomment-1268791630

### DIFF
--- a/src/api/plugins.js
+++ b/src/api/plugins.js
@@ -168,7 +168,7 @@ module.exports = function buildJSONApi(gridsome) {
                 ? new URL(plugin.logo_url, baseurl)
                 : undefined;
               const docs = new URL(
-                path.join(pluginType.path, plugin.name, plugin.variant),
+                path.join(pluginType.path, `${plugin.name}--${plugin.variant}`),
                 baseurl
               );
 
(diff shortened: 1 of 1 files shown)


src/api/plugins.js
```diff
@@ -165,23 +165,23 @@ module.exports = function buildJSONApi(gridsome) {
             result.data.data.edges.map(async ({ node: plugin }) => {
               // Clean up plugin object, create full log_url, add docs url
               const logoUrl = plugin.logo_url
-                ? path.join(baseurl, plugin.logo_url)
+                ? new URL(plugin.logo_url, baseurl)
```
edgarrmondragon, line 168: I think gridsome is pre-processing the logo images so these URLs are not valid. For example: https://deploy-preview-877--meltano-hub.netlify.app/assets/logos/extractors/stackexchange.png.

That seems unrelated to the base URL issues, though.

edgarrmondragon (approved):

src/api/plugins.js
rwfeather: Yeah you're right. #878 should resolve the issue

PR #878, opened 2022-10-05
Move whole assets folder into `static`

rwfeather:
Gridsome is frustratingly unable to include the top-level `assets` dir into the static bundle. It has been working so far because there's a built-in vue component to do dynamic image processing at build time.

In order to serve things out of the `assets` dir, it needs to be moved under `static`.

This was affecting:

- The `logo_url` property in our JSON API (the urls were correct but they reached a 404)
- The embedded images in a few plugins' markdown documentation


edgarrmondragon (approved):
:shipit:

PR #890, opened 2022-10-11
chore: update logos to remove wordmark

tayloramurphy:
This PR updates some of the logos to remove the wordmark.


WillDaSilva (approved):
I was initially worried about the decrease in accessibility (to those unfamiliar with these logos), but I think everywhere these logos appear they are accompanied by the text that this update removes, so this update really just seems to make things cleaner. :+1:

afolson:
@tayloramurphy Can you use the dbt icon with the trademark symbol? It's in their asset library: https://www.getdbt.com/brand-guidelines/ && https://drive.google.com/drive/folders/1lCP2GbKJud6zk_DZAivP8DV-7PLUYxXV

I couldn't find brand guidelines for Airflow or Metabase so those are probably fine as is. Thanks!

PR #895, opened 2022-10-13
change: Mob session 2022-10-13

aaronsteers:


_data/maintainers.yml
```diff
@@ -1,4 +1,4 @@
-a-rusi:
+a-rusi:
```
aaronsteers, line 1:
```suggestion
a-rusi:
```

revert

tayloramurphy:
Looks good but it looks like there's conflicts on the images.

tayloramurphy (approved):

_data/meltano/extractors/tap-mysql/transferwise.yml
```diff
@@ -19,38 +19,66 @@ settings:
...
-  description: The MySQL database name.
+  description: The MySQL/MariaDB database name.
+- name: cursorclass
+  label: Cursor Class
+  value: pymysql.cursors.SSCursor
+  description: Cursor class used by PyMYSQL.
+- name: server_id
+  label: Server ID
+  kind: integer
+  value: <random int>
```
aaronsteers, line 45: In order to let the tap derive the random ID here, I think we need to remove this line. Otherwise, I think '<random int>' will fail conversion to type 'int'.

```suggestion
```

_data/meltano/extractors/tap-github/singer-io.yml
```diff
@@ -28,7 +28,7 @@ settings:
     kind: date_iso8601
     description: Defines how far into the past to pull data for the provided repositories.
 capabilities:
-  - properties
+  - catalog
```
aaronsteers, line 31: We discovered during the mob session that we can't always trust the --help output because it may be calling the common singer library without actually handling all declared CLI options.

So I dove into the docs at https://github.com/singer-io/tap-github and also I checked the source code. In fact, this tap does expect properties and does not appear to support catalog.


```suggestion
  - properties
```

afolson (approved):

PR #904, opened 2022-10-13
fix: Hide capabilities section when no capabilities are defined.

### LINKED ISSUE #901: Don't render "Capabilities" section for loader plugins that don't define capabilities
Many loaders have no capabilities defined. In those cases, we currently render the normal summary/disclaimer text about capabilities but then have no capabilities listed below it. This can be confusing and should be removed.

cjohnhanson:
Closes #901


src/components/PluginCapabilitiesSection.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div>
     <p class="text-3xl py-4" id="capabilities">Capabilities</p>
-    <span class="space-y-3" v-if="capabilities">
+    <span class="space-y-3" v-if="len(capabilities) > 0">
```
tayloramurphy, line 4:
```suggestion
    <span class="space-y-3" v-if="capabilities && capabilities.length">
```

Based on how settings is done I think this is the way?

tayloramurphy (approved):

tayloramurphy:
Looks good on https://deploy-preview-904--meltano-hub.netlify.app/transformers/dbt-bigquery vs https://hub.meltano.com/transformers/dbt-bigquery/

edgarrmondragon (approved):
lgtm :shipit:

PR #928, opened 2022-10-14
add description_rendered

tayloramurphy:
Closes https://github.com/meltano/hub/issues/930

### DIFF
--- a/src/templates/Plugins.vue
+++ b/src/templates/Plugins.vue
@@ -289,7 +289,8 @@ query Plugins($path: String!, $name: String!) {
   plugins: plugins(path: $path) {
     id
     description
-    description_rendered
+    definition
+    definition_rendered
     label
     name
     path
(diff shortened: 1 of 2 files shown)


src/templates/Plugins.vue
```diff
@@ -76,11 +76,15 @@
...
+              <span 
+                class="prose"
+                v-if="$page.plugins.pluginType === 'utility'"
+                v-html="$page.plugins.definition_rendered"></span
               >
-              <span v-if="$page.plugins.pluginType === 'file'">
-                {{ $page.plugins.definition }}</span
+              <span 
+                class="prose"
+                v-if="$page.plugins.pluginType === 'file'">
```
edgarrmondragon, line 86:
```suggestion
                v-if="$page.plugins.pluginType === 'file'"
```

cjohnhanson (approved):
Looks like linting is failing, but other than that this looks good to me :ship:

tayloramurphy:
I'm not entirely happy with this but it's a stable iteration and better than having nothing. will create a follow-up issue

PR #935, opened 2022-10-17
Hub bot - Missing Plugins

pnadolny13:
Closes https://github.com/meltano/squared/issues/392

This also bumps the search bar text to "350+" because were closer to 400 after this.

- Adds 40 new extractors
- Adds 1 extractor non-default variant
- Adds 2 loaders

All of these meet this criteria:
- not a fork
- anonymous telemetry shows usage
- has activity in the last 12 months
- greater than 7 days old (I dont think there were any close to 7 days, I'd guess the youngest is 1 month old)
- greater than 1 commit
- I manually viewed all of them
- I ran the utility CLI described below so install and `--help` worked

I created a utility CLI as part of this https://github.com/pnadolny13/hub-utils that helps streamline some of the repetitive stuff, which does:
  - builds boilerplate definition
  - adds maintainer if not exists
  - adds default variant entry
  - installs
  - runs `meltano invoke tap-x --help` to check for valid install and execution
  - if its SDK based we can extract settings, group validations, capabilities, etc. using `--about --format=json`
  - Intended to be a helper that needs to be human reviewed before committing

@aaronsteers related to https://github.com/meltano/hub/pull/893#issuecomment-1278509719. I closed that PR because the blind automated approached seemed wrong, especially since a definition without settings isn't very useful so I've pivoted to an accelerator/automation approach. The process (at least for now) can be a person evaluating them with the helper CLI that accelerates adding.


tayloramurphy (approved):
I pushed a small change to sort order on the default and update the Meltano logo. Otherwise this is SUPER impressive. Nice work @pnadolny13

afolson:
This looks great. Thanks @pnadolny13 🎉

PR #940, opened 2022-10-19
feat: Build API files separate from site deployment

edgarrmondragon:
Partial solve towards: https://github.com/meltano/internal-general/issues/444

- [x] Build API files separate from Gridsome build
- [x] Upload them as GitHub artifacts so the can be used in downstream jobs (e.g. copy to S3)
- [ ] Copy files to S3 bucket
- [x] Parametrize base hub URL in logo and docs URLs using PR number
- [ ] Parametrize base API URL in "ref" URLs in the `/index` endpoints


rwfeather (approved):
lgtm, we were gonna have to divorce this from the netlify build at some point :)

Once we've completed the migration we'll want to remove the API generation from the gridsome site build.

edgarrmondragon:
@aaronsteers I'll merge this after the checks pass, if that's OK. I _think_ we have issues for the unchecked items from the PR description.

Or should I wait to have those in place here?

PR #952, opened 2022-10-20
feat: Add test dispatcher workflow

### LINKED ISSUE #917: On new PRs, do a diff against `main` and (re)test any files changed under the `_data/` dir
We have a pattern now for testing plugins - or at least for executing them and scanning their metadata.

In new PRs, we could diff against `main` any changed files in the `_data/` directory, and pass those plugins' to something similar to the slash command processor.

Perhaps in this use case we don't actually need to push comments back to the PR, as a CI job output may be completely sufficient - especially if we use the relatively new feature of providing markdown result rendering on CI jobs.

WillDaSilva:
Example usage: https://github.com/WillDaSilva/meltano-hub/pull/1/

It was tested on a fork because I had to use `pull_request_target`, which runs in the context of the main branch, and so won't have an effect until this PR is merged into `main`.

Note that if it had previously commented about a plugin (as identified by its name and variant).

The linked example shows it printing the pip url, but it no longer does this. The variant is printed instead.

Closes #917

### DIFF
--- a/.github/workflows/test_dispatcher.yml
+++ b/.github/workflows/test_dispatcher.yml
@@ -46,6 +46,8 @@ jobs:
     steps:
       - name: Checkout the repository
         uses: actions/checkout@v3
+        with:
+          ref: "refs/pull/${{ github.event.pull_request.number }}/merge"
 
       - name: Install yq
         env:
--- a/.github/workflows/test_dispatcher.yml
+++ b/.github/workflows/test_dispatcher.yml
@@ -1,0 +1,81 @@
+
+name: Plugin 'Test' Command Dispatcher
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+    paths:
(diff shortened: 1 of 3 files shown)


.github/workflows/test_dispatcher.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      - id: set-matrix
+        run: echo "matrix={\"changed_file\":${{ steps.filter_files.outputs.filtered_files }}}" >> "$GITHUB_OUTPUT"
+
+  dispatch_tests:
+    runs-on: ubuntu-latest
+    needs:
+      - get_changed_files
+    strategy:
+      matrix: ${{ fromJson(needs.get_changed_files.outputs.matrix) }}
```
aaronsteers, line 43: 🤯

rwfeather (approved):

aaronsteers (approved):

PR #961, opened 2022-10-25
Tap twitter variants

pnadolny13:
- the existing tap variant is 5 years old and fails to install see https://github.com/bernietai/tap-twitter/issues/1. The auth flow is pretty confusing too.
- theres a new SDK based variant that exists to I made that the default. It doesnt pull as much data but it worked for me and I think we'd rather have someone succeed and need to contribute more streams to fit their use case than fail and give up.

### DIFF
--- a/_data/default_variants.yml
+++ b/_data/default_variants.yml
@@ -356,7 +356,7 @@ extractors:
   tap-twinfield: yoast
   tap-twitter-ads: singer-io
   tap-twitter-streams: pbegle
-  tap-twitter: bernietai
+  tap-twitter: voxmedia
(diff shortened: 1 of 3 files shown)


WillDaSilva:
> the bot ran tests on `tap-postgres/tap-db2` even though I didnt make any changes to those :thinking:. Could it be a bug with the test bot?

@pnadolny13 Yeah, seems so. `tap-db2` was updated on `main` (see https://github.com/meltano/hub/pull/958). The bot runs on the PR merge commit, so maybe it's picking up changes from the target branch. I'll open an issue.

tayloramurphy (approved):

PR #964, opened 2022-10-25
Setting round 4

pnadolny13:
- More legacy singer settings
- Add a few non-default SDK variants for awareness


tayloramurphy (approved):
👏 nice work

PR #971, opened 2022-10-26
fix: Dispatch PR plugin tests for forks

### LINKED ISSUE #969: bug: PR auto test fails on fork PRs
https://github.com/meltano/hub/actions/runs/3326476744/jobs/5500226517

I think the test job is failing because the PR is from a fork that it cant access. I'm not sure how we can handle this better other than excluding these tests on forks.

WillDaSilva:
Example of it working with a PR from MeltyBot's fork to my fork: https://github.com/WillDaSilva/meltano-hub/pull/2

Closes #969

### DIFF
--- a/.github/workflows/test_dispatcher.yml
+++ b/.github/workflows/test_dispatcher.yml
@@ -20,6 +20,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
 
       - name: Get changed files
         id: changed_files
@@ -47,7 +49,7 @@ jobs:
       - name: Checkout the repository
         uses: actions/checkout@v3
         with:
-          ref: "refs/pull/${{ github.event.pull_request.number }}/merge"
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Install yq
         env:


WillDaSilva:
After this has been merged, I'll delete MeltyBot's fork of the Hub repository.

pnadolny13:
@WillDaSilva this looks good to me!

The only question I'd have is if there are any risks of us automatically running CI jobs on forks, not saying that there is but that I actually dont know what the risks are (you probably know more than me). In the past I asked about how we can run CI tests for taps in meltanolabs and I heard some people say that we've skipped tests that are from forks, or require a manual trigger from someone on the team. I definitely want to have the tests run on forks because the majority of the failures will come from community contributors who are less familiar with the schemas/structure of files but I also wouldnt want us to expose ourselves to risk.

What do you think? I'd probably defer this review back to engineering cc @aaronsteers @cjohnhanson @rwfeather .

WillDaSilva:
> The only question I'd have is if there are any risks of us automatically running CI jobs on forks [...]

@pnadolny13 The test dispatch job runs on [`pull_request_target` events](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target), which are executed in the context of the base branch of the pull request. This prevents users from changing the workflow definition to do malicious things, as they'd have to get those workflow updates merged into e.g. `main` in our repository before they could actually run it.

See also https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

We should consider any files from a fork to be untrusted input, and not run any code from a fork without verifying it first. Fortunately the test dispatch need only parse some yaml files, which should not pose any considerable risk. Then we run the test command, which may have vulnerabilities (not that I've thought of any yet), but malicious users can already leave comments in our repository to trigger the test command, so there's no additional risk from letting the test dispatcher do it.

rwfeather (approved):

PR #974, opened 2022-10-27
Updates extractors tap-mailchimp (singer-io), and make default

pnadolny13:
Adds settings to singer-io variant and makes it the default. The [singer-io variant](https://github.com/singer-io/tap-mailchimp) has had updates in the last 12 months and I see a bunch of new PRs so its active vs the [lovepopcards variant](https://github.com/lovepopcards/tap-mailchimp) (currently default) that hasnt had a commit since 2017. 

@tayloramurphy any idea why we might have chosen to make lovepopcards the default over singer-io?

### DIFF
--- a/_data/default_variants.yml
+++ b/_data/default_variants.yml
@@ -222,7 +222,7 @@ extractors:
   tap-loopreturns: loopreturns
   tap-lotr: mattarderne
   tap-maestroqa: pathlight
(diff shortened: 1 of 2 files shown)


tayloramurphy:
@pnadolny13 I don't recall how or why we picked that as the default. Thanks for the PR!

Issue #950, opened 2022-10-20
Yaml Format CI Test


pnadolny13:
Following https://github.com/meltano/hub/issues/949 we should add a little CI test that makes sure new additions and updates conform to the style guide that we decide on.

edgarrmondragon:
I've used [yamllint](https://yamllint.readthedocs.io/en/stable/configuration.html) before with some success

PR #979, opened 2022-10-28
ci: Initial pass at linting YAML files

edgarrmondragon:
Full `yamllint` rules reference: https://yamllint.readthedocs.io/en/stable/rules.html#

Closes https://github.com/meltano/hub/issues/950


edgarrmondragon:
@pnadolny13 this is a first pass at getting linting setup for YAML files.

It adds nice annotations: https://github.com/meltano/hub/pull/979/files#annotation_5518730485.

I'm not sure if we want to apply fixes for _all_ files in one big PR, but I can look into that. Otherwise this approach is incremental and should only check files touched by the PR.

Wdyt?

pnadolny13:
> @pnadolny13 this is a first pass at getting linting setup for YAML files.
> 
> It adds nice annotations: https://github.com/meltano/hub/pull/979/files#annotation_5518730485.
> 
> I'm not sure if we want to apply fixes for _all_ files in one big PR, but I can look into that. Otherwise this approach is incremental and should only check files touched by the PR.
> 
> Wdyt?

@edgarrmondragon this is great! Maybe having a blurb in the [CONTRIBUTING.md](https://github.com/meltano/hub/blob/main/CONTRIBUTING.md) around how to test locally before pushing changes would be helpful also.

Yeah I like the idea of one giant PR to update all of them. I didnt see a `fix` type command like sqlfluff has. Is there a way to have it update the file where it can? I guess we'd have to write some scripts for that if not.

edgarrmondragon:
> @edgarrmondragon this is great! Maybe having a blurb in the [CONTRIBUTING.md](https://github.com/meltano/hub/blob/main/CONTRIBUTING.md?rgh-link-date=2022-10-31T16%3A38%3A53Z) around how to test locally before pushing changes would be helpful also.

Sounds good. I've added pre-commit, which can be run locally.

> Yeah I like the idea of one giant PR to update all of them. I didnt see a `fix` type command like sqlfluff has. Is there a way to have it update the file where it can? I guess we'd have to write some scripts for that if not.

I gave [`opt-nc/yamlfixer`](https://github.com/opt-nc/yamlfixer) a try, but it doesn't fix key ordering and isn't great at detecting multi-line strings that overflow the line length limit 😕.

rwfeather (approved):
lgtm 👍

.prettierignore
```diff
@@ -5,4 +5,6 @@ src/.temp
 _data
 assets
 schemas
-singer
\ No newline at end of file
+singer
+.pre-commit-config.yaml
+.yamllint
```
cjohnhanson, line 10: I think this should resolve the failing linting test?
```suggestion
.yamllint.yaml
```

cjohnhanson (approved):
Looks good, pending that one failing check.

.prettierignore
cjohnhanson: I went ahead and merged this and it looks like it did the trick.

edgarrmondragon:
@cjohnhanson thanks for the linter update!

PR #999, opened 2022-11-03
Filter variant list for current plugin type

### LINKED ISSUE #898: bug: Plugins showing duplicate variants when a plugin of a different type exists with the same name
https://hub.meltano.com/extractors/tap-zendesk/ is showing meltano as a default variant but that's coming from the `transforms`

rwfeather:
Closes #898


tayloramurphy (approved):
Looks good in the UI!

PR #1008, opened 2022-11-15
add more details to tap-slack about setting up app

pnadolny13:
Related to slack thread https://meltano.slack.com/archives/CMN8HELB0/p1668441644910859 where 
a user had issues syncing data because they didnt have proper permissions. Just adds some detail from the repo README about creating an App and what score is needed.


sbalnojan (approved):
Looks great, thanks for the quick response the user is still struggling with it!

I am wondering however about one thing: required settings. (API key is required here) Do we have a default place to put these? 

And a meta question: The list of settings and required settings should be automatically parsable from the plugin itself, shouldn't it? Is there more to it than it appears to be? Or are we duplicating work here?

tayloramurphy (approved):
LGTM! I also updated the logo to remove the wordmark.

pnadolny13:
>I am wondering however about one thing: required settings. (API key is required here) Do we have a default place to put these?

@sbalnojan yep we have [settings_group_validation](https://docs.meltano.com/reference/plugin-definition-syntax#settings_group_validation) that defines what keys are required.

>And a meta question: The list of settings and required settings should be automatically parsable from the plugin itself, shouldn't it? Is there more to it than it appears to be? Or are we duplicating work here?

I'm not sure I follow exactly, can you say more? We define settings and required settings in the plugin definition so Meltano knows what to request from the user and the hub has detail for the plugin pages, most singer taps/targets executables dont have a way to share their metadata although SDK based versions do now using `--about`.

Issue #1018, opened 2022-11-18
Freshdesk Configuration Invalid


visch:
From Florian on Slack https://meltano.slack.com/archives/C01TCRBBJD7/p1668714398069059

```
Plugin configuration is invalid
Catalog discovery failed: command ['/home/lfadmin/meltano-projects/freshdesk_tst/.meltano/extractors/tap-freshdesk/venv/bin/tap-freshdesk', '--config', '/home/lfadmin/meltano-projects/freshdesk_tst/.meltano/run/tap-freshdesk/tap.9b3d55e9-5861-43bd-8167-03d93eca7f86.config.json', '--discover'] returned 2 with stderr:
 usage: tap-freshdesk [-h] -c CONFIG [-s STATE]
tap-freshdesk: error: unrecognized arguments: --discover
```

My best bet is we need to update capabilities here https://github.com/meltano/hub/blob/main/_data/meltano/extractors/tap-freshdesk/singer-io.yml#L32 to 

```
capabilities:
- state
```

pnadolny13:
@visch thanks for logging it I created a PR to remove the other capabilities.

PR #1020, opened 2022-11-18
fix freshdesk capabilities

pnadolny13:
Closes https://github.com/meltano/hub/issues/1018

I've actually never seen a tap that doesnt support catalog/discover. I wonder if theres implications on the meltano side. I know selection rules wouldnt work in this case. @edgarrmondragon do you know?

cc @visch


edgarrmondragon:
> I know selection rules wouldnt work in this case. @edgarrmondragon do you know?

@pnadolny13 That's right. The `select`, `select_filter`, `metadata` and `schema` extras wouldn't have any effect.

PR #1022, opened 2022-11-18
Add generic issue template

afolson:
@tayloramurphy @pnadolny13 We didn't have a generic bug report issue template so I figured it would be good to make one so that we can collect the right info from folks. Will this work?

@rwfeather @cjohnhanson Would love to get your thoughts on this to make sure this captures info you might need as well.

See https://meltano.slack.com/archives/C01UGBSJNG5/p1668779438979179 for context on this proposal.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,40 @@
+---
+name: Bug report
+about: Report a potential bug in the MeltanoHub
+title: "\U0001F41B BUG: `Brief description of the problem` "
+labels: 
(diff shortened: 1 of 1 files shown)


tayloramurphy:
@afolson to get this to work like the others, I think it needs to be a YAML file so it can render like this https://github.com/meltano/hub/blob/main/.github/ISSUE_TEMPLATE/new_tap.yml content-wise I think it looks good though!

cjohnhanson (approved):
@afolson this looks good to me, as long as we have device and browser that should help us reproduce and debug so I think this gives us all the info we need.

Issue #1021, opened 2022-11-18
Reformat All Yaml To Pass Linting


pnadolny13:
After https://github.com/meltano/hub/pull/979 was merged we're getting CI test failures due to existing formatting violations. We should limit the amount of work that everyone has to do piece meal by doing a giant reformat of all yaml files at once.

cc @edgarrmondragon

edgarrmondragon:
Agreed, it's a shame that https://github.com/opt-nc/yamlfixer doesn't do what we need since it's supposed to work in tandem with yamllint.

PR #1023, opened 2022-11-18
Script to reformat yaml to pass linter tests

pnadolny13:
Related to https://github.com/meltano/hub/issues/1021

This PR adds a script that assists in getting our yaml files to conform to the new yaml formatting CI tests we added in https://github.com/meltano/hub/pull/979. The script will iterate all yml files and fix them for our one time backfill and it also accepts a file path so moving forward it can help if tests fail.

As part of this I wasnt able to remove trailing spaces or split long lines so I set those rules to be ignored for now. The issue is that when I edit a string value it writes back to the yaml file wrapped in double quotes vs the nice pipe/indent syntax like https://github.com/meltano/hub/blob/4eafa14a013b762342d2c5766465b22205c4d999/_data/meltano/loaders/target-bigquery/jmriego.yml#L16. I have a mechanism to split lines but I commented out the code for now until we figure out how to keep the clean formatting or accept the double quotes. If anyone has more experience with ruamel and knows how to do this 🙏 please provide suggestions.

I ran this over our repo multiple times while testing but didnt want to explode this PR while we review the actual code changes. I can open a follow on PR where I actually run it to reformat hundreds of files 😅 .


tayloramurphy (approved):
I've got something funky going on with my Python install locally so I can't really test this right now. Happy to approve if you're happy with the output.

edgarrmondragon (approved):

PR #1026, opened 2022-11-21
reformat all yaml files

pnadolny13:
Closes https://github.com/meltano/hub/issues/1021

This is a follow on to https://github.com/meltano/hub/pull/1023 that actually runs the yaml fix script over all the files. See https://github.com/meltano/hub/pull/1023 for more details on what it does but its mostly just sorting keys right now. It should at least help solves the majority of the issues that the CI tests will error on.


tayloramurphy (approved):
@pnadolny13 seems like the tests are happy and the hHub renders. LGTM!

PR #1031, opened 2022-11-22
fix: Fix built API urls and document new script

edgarrmondragon:
- Ensures the `/plugins/` part of the URL is set.
- Writes files to the `<output-dir>/plugins` directory.
- Documents the script usage


rwfeather (approved):
lgtm 🚀

PR #1033, opened 2022-11-23
add settings for tap/target postgres and tap snowflake

pnadolny13:


pnadolny13:
@edgarrmondragon any idea what the yamllint CI [test failure](https://github.com/meltano/hub/actions/runs/3533555055/jobs/5929326193) means? 

```
  Error: Unable to locate a common ancestor between main and e5a824ebd3aa3856cdf2ee1893525c2b46defe13
  Error: Process completed with exit code 1.
```

pnadolny13:
@tayloramurphy I was planning to add target-snowflake but [it failed](https://meltano.slack.com/archives/C04BG3978NT/p1669218742033839) during my hub-util add script so I excluded it for now. 

>It'd also be worth hiding https://hub.meltano.com/loaders/target-snowflake--meltano/ since we've pointed folks to the transferwise variant already.

I think that might be a bug in the rendering since it already has the `hidden` attribute https://github.com/meltano/hub/blob/01d37aa1bd4ff018c0912dc4af6b433e74471032/_data/meltano/loaders/target-snowflake/meltano.yml#L4 but it shows up still.

tayloramurphy (approved):

edgarrmondragon:
> @edgarrmondragon any idea what the yamllint CI [test failure](https://github.com/meltano/hub/actions/runs/3533555055/jobs/5929326193) means?
> 
> ```
>   Error: Unable to locate a common ancestor between main and e5a824ebd3aa3856cdf2ee1893525c2b46defe13
>   Error: Process completed with exit code 1.
> ```

@pnadolny13 I'm not sure I follow [this explanation](https://github.com/tj-actions/changed-files/issues/809#issuecomment-1319162119)

> High-level description: This bug occurs when the fork point and the last commit of the base branch differ, resulting in errors finding the merge-base as the checkout action merges the tip of the target branch to the PR branch.
> 
> To prevent this error going forward a pre-check to determine the validity of merge-base would be used falling back to the last commit of the target branch where necessary.

But it seems that bumping the action would fix the issue. I'll start a PR.

PR #1034, opened 2022-11-23
ci: Add dependabot updates for actions and Python deps

edgarrmondragon:
See https://github.com/meltano/hub/pull/1033#issuecomment-1325438057


pnadolny13 (approved):
@edgarrmondragon LGTM!

PR #1035, opened 2022-11-23
Add Google Tag Manager

DouweM:


rwfeather (approved):
🙌

cjohnhanson:
@DouweM (cc: @rwfeather) I'm not super familiar with how Google Analytics sends this data, but I still see the old ID when I check the network activity in the deploy preview: 
Screenshot 2022-11-23 at 4 58 00 PM

Would this indicate that the config change isn't working as expected?

cjohnhanson (approved):
Tentatively approving so we can get this out there, but definitely want to call out: https://github.com/meltano/hub/pull/1035#issuecomment-1325748787

I could just be misunderstanding how the google analytics client sends this data, but I'm seeing the `G-NSVMGNSZKN` ID in the query params for calls to Google Analytics' `/collect` endpoint in the deploy preview.

DouweM:
@cjohnhanson Kudos for actually checking in the browser! The way Google Tag Manager works is that it allows you to manage all your tags in their interface, that are then all injected dynamically using the single GTM tag you actually put in your website. As one of those tags we've configured is still Google Analytics with the original ID, it's totally expected that you'd still see that collect request.

PR #1041, opened 2022-11-23
bugfix: empty command list

pnadolny13:
This fixes a bug where the commands section is included even if the plugin doesnt have any:

Screen Shot 2022-11-23 at 1 38 14 PM

See "Commands" section of https://hub.meltano.com/extractors/tap-postgres. Now it makes sure that there are elements in the list before populating the sections header.

This works but let me know if theres a more javascripty way to do the "not null and not empty" conditional.


pnadolny13:
Closing since this is a duplicate of https://github.com/meltano/hub/pull/1042

PR #1042, opened 2022-11-25
Do not render `Commands` section for plugins that do not define any `commands`

ReubenFrankel:
Small QOL change for plugins like [`tap-auth0`](https://hub.meltano.com/extractors/tap-auth0/):

image


tayloramurphy (approved):
I like this change. Makes it cleaner. Thanks @ReubenFrankel

pnadolny13 (approved):

PR #1043, opened 2022-11-25
Add missing capabilities and setting labels for `tap-spotify`

ReubenFrankel:
[tap-spotify](https://hub.meltano.com/extractors/tap-spotify/) settings had no labels, causing them to be displayed in an odd way.

image

Also updated `capabilities` for good measure.


tayloramurphy (approved):
Thanks for the PR!

Issue #1045, opened 2022-11-25
Test dispatcher tries to install plugins without a `pip_url`


edgarrmondragon:
Testing plugin `clickhouse` (`clickhouse` variant):
> [Starting test job...](https://github.com/meltano/hub/actions/runs/3549703768)
> ...
> Job completed.

melty-bot % pipx install null
creating virtual environment...
installing null...
No apps associated with package null or its dependencies. If you are
attempting to install a library, pipx should not be used. Consider using pip
or a similar tool instead.

_Originally posted by @github-actions[bot] in https://github.com/meltano/hub/issues/1044#issuecomment-1327727662_

PR #1046, opened 2022-11-25
fix: Skip test for plugins without a `pip_url`

edgarrmondragon:
Closes https://github.com/meltano/hub/issues/1045

### DIFF
--- a/.github/workflows/test_dispatcher.yml
+++ b/.github/workflows/test_dispatcher.yml
@@ -62,13 +62,14 @@ jobs:
       - name: Get the name and pip url
         run: |
           name="$(basename "$(dirname "${{ matrix.changed_file }}")")"
-          pip_url="$(./yq '.pip_url' "${{ matrix.changed_file }}")"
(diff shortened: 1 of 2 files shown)


edgarrmondragon:
Can't run the updated workflow in a PR:

> * With `on: pull_request_target` as trigger, secrets can be accessed, but the workflow of the base branch is used (so the version trusted by you). Any changes done to `.github/workflows/` by the PR are ignored, which should prevent malicious users from leaking secrets. Note that also the code of the base branch is used unless you explicitly checkout the PR code:

https://github.com/orgs/community/discussions/27084#discussioncomment-3254544

edgarrmondragon:
Confirmed that the skip works in my fork: https://github.com/edgarrmondragon/hub/actions/runs/3550350798/jobs/5963649414

tayloramurphy (approved):

tayloramurphy:
Love this - thanks Edgar!

PR #1055, opened 2022-11-30
exclude tap-slack new admin stream by default

pnadolny13:
Related to https://github.com/MeltanoLabs/tap-slack/pull/14

Theres a new stream that requires admin credentials where as the existing streams do not. My thought was that we could allow that stream to be added to the tap but by default exclude it from syncs. If a user wants to add it they can override the default select criteria but it wont affect normal users.

If we like this idea I can also add a blurb to one of the free text sections like `usage` to describe that the stream is available for admin users.

@tayloramurphy @edgarrmondragon what do you think?

### DIFF
--- a/_data/meltano/extractors/tap-slack/meltanolabs.yml
+++ b/_data/meltano/extractors/tap-slack/meltanolabs.yml
@@ -55,6 +55,9 @@ prereq: |
   curl -X GET -H 'Authorization: Bearer <Your Token>' https://slack.com/api/conversations.list?exclude_archived=False
   ```
 repo: https://github.com/MeltanoLabs/tap-slack
+select:
+- '*.*'
+- '!integration_logs.*'
 settings:
 - description: Your Slack App API key. See the Prerequisites section above for more
     details on obtaining it.


edgarrmondragon (approved):
This is a sensible approach, and I was not aware but it seems we're already doing this in other taps 👍

tayloramurphy:
@pnadolny13 makes sense to me. Feel free to add the comment to the usage or next_steps sections and merge.

aaronsteers:
@pnadolny13 - We also have some precedent in tap-gitlab where a config option would enable/disable streams that the user would not have access to or do not have license to.

If there might be others that need admin access, a `include_admin_streams` or similar. On the Gitlab side, there were a bunch of streams that were only available in Ultimate, if I remember correctly.

afolson (approved):

pnadolny13:
@aaronsteers I like that idea better, especially after realizing that when a user changes the default selection criteria they need to remember to exclude this stream. I feel like that would cause people to think theres a bug if it fails. I'll make this suggestion on the tap PR.

Issue #185, opened 2022-01-20
Add missing taps


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/hub/-/issues/189

Originally created by @pnadolny13 on 2022-01-20 15:03:10

---

Some taps that arent listed that we should consider adding:

- [ ] https://github.com/albhu/tap-buildium/tree/master/tap_buildium 
- [x] https://github.com/manuelapilongo/tap-brillium (SDK)
- [ ] ~~https://github.com/tharsheblows/tap-mjjwordpressrest (SDK)~~
- [ ] https://github.com/decentralgames/tap-dg-ice (SDK)
- [x] https://github.com/dwallace0723/tap-fabdb
- [x] https://github.com/birdiecare/tap-cqc-org-uk
- [x] https://github.com/potloc/tap-wrike
- [ ] ~~https://github.com/weslst/tap-crunchbase~~
- [x] https://github.com/evelte/tap-getpocket
- [x] https://github.com/Datateer/tap-clickcast
- [x] https://github.com/manuelapilongo/tap-aws-cognito (SDK)
- [x] https://github.com/dluftspring/tap-hellobaton
- [ ] ~~https://github.com/Blue-Rabbit-Vet/tap-cscart~~
- [ ] https://github.com/morrislaptop/tap-gocardless (other variant exists too)
- [ ] https://github.com/MichaelCannucci/tap-callrail (SDK) - needs <3.9
- [x] https://github.com/singer-io/tap-amazon-sp
- [x] https://github.com/storebrand/tap-sharepointsites
- [x] https://github.com/dan-ladd/target-salesforce (SDK)
- [x] https://github.com/Widen/tap-gainsightpx
- [ ] https://github.com/BuzzCutNorman/target-mssql (SDK)
- [x] https://github.com/BuzzCutNorman/tap-mssql (SDK)
- [ ] https://github.com/barloc/tap-dataai (SDK) - type error, https://github.com/barloc/tap-dataai/issues/1
- [x] https://github.com/epoch8/tap-paypal (SDK)
- [x] https://github.com/epoch8/tap-hotjar (SDK)
- [x] https://github.com/degreed-data-engineering/tap-datadog  (SDK) - https://github.com/degreed-data-engineering/tap-datadog/issues/5
- [ ] https://github.com/splitio/tap-datadog
- [x] https://github.com/iomete/singer-target-iomete
- [x] https://github.com/robby-rob-slalom/tap-podbean (SDK)
[...]

PR #1064, opened 2022-12-05
SDK based variants

pnadolny13:
Adds 21 new SDK based taps/targets continuing to burn down the list from https://github.com/meltano/hub/issues/185.

New:
- 1 target
- 15 extractors

Alternative Variants:
- 5 extractors total. 1 is now the new default (tap-indeed)

Bugfix: The CI build preview was failing because some of these new plugins dont have metrics. Adding a check to make sure that metrics isnt null before using it looks to fix it.


tayloramurphy:
@pnadolny13 I'm able to find a few better images for some of the connectors - redis and aftership in particular.

Also, if you're able to find stacked logos w/ wordmarks instead of the horizontal ones that'd be nice.

tayloramurphy (approved):

tayloramurphy (approved):

PR #1071, opened 2022-12-07
more spreadsheets anywhere details

pnadolny13:
This tap gets recommended all over the place and I found it hard to use.
- I copied the tables keys and descriptions so you dont necessarily have to leave the hub to understand how to use it
- Added a section to give a few simple functioning examples that I've gotten working, in case they help others. We can grow that section over time if we want

### DIFF
--- a/_data/meltano/extractors/tap-spreadsheets-anywhere/ets.yml
+++ b/_data/meltano/extractors/tap-spreadsheets-anywhere/ets.yml
@@ -31,6 +31,32 @@ settings:
 - description: |
     An array holding json objects that each describe a set of targeted source files.
 
+    Each object in the 'tables' array describes one or more CSV or Excel spreadsheet files that adhere to the same schema and are meant to be tapped as the source for a Singer-based data flow.
+    The available keys are:
+
+    - path: A string describing the transport and bucket/root directory holding the targeted source files.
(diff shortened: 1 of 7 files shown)


_data/meltano/extractors/tap-spreadsheets-anywhere/ets.yml
```diff
@@ -50,10 +76,48 @@ settings:
...
+
+  ```
+    config:
+      tables:
+      - path: s3://<your-bucket-name>
+        format: json
+        key_properties: [user_id]
+        name: user_names
+        start_date: '2020-01-01T00:00:00Z'
+        pattern: <prefix>/user_names.json
```
sbalnojan, line 114: I think this will be hard to understand (that "pattern" refers to a regex pattern is already hard to grasp). I would put in an actual prefix use an actual example...

sbalnojan (approved):

sbalnojan (approved):

_data/meltano/extractors/tap-spreadsheets-anywhere/ets.yml
pnadolny13: @sbalnojan good point! I updated it to an example string instead of a placeholder.

PR #1072, opened 2022-12-07
fix mysql test feature

pnadolny13:
Related to an issue in the singer-io tap-mysql repo https://github.com/singer-io/tap-mysql/issues/154. I confirmed that the `config test` command fails even with valid configs set.

If I set a default selection criteria and metadata then all is good. I feel like this is a bad idea since it auto selects everything and discovery might be slow. I tried defaulting to `!*.*` selection so defaulted to nothing but that fails `config test` still.

What exactly is the `config test` command running? Should we have a way to disable that `config test` command block for certain plugin pages?

@edgarrmondragon @tayloramurphy @sbalnojan what do you all think?


tayloramurphy:
@pnadolny13 this seems reasonable. I would just recommend putting some warnings / alerts in the documentation for this change too.

pnadolny13:
@tayloramurphy yeah that sounds good. I actually figured out that we dont need selection criteria, just the replication-method metadata. I'll still add a note that its defaulting to FULL_TABLE replication now.

edgarrmondragon:
> What exactly is the `config test` command running? Should we have a way to disable that `config test` command block for certain plugin pages?

@pnadolny13 It invokes the tap, full with discovery, etc. It then exits as soon as it sees the first `RECORD` message.

Otherwise +1 to Taylor's comment above.

pnadolny13:
@rwfeather @edgarrmondragon @cjohnhanson do any of you know whats up with the failing build here? I get the same thing locally and couldnt figure out where its coming from. I confirmed that not using the star allowed it to build. I wonder if something with gridsome/the graphql package hates that syntax 🤔 .

edgarrmondragon:
> @rwfeather @edgarrmondragon @cjohnhanson do any of you know whats up with the failing build here? I get the same thing locally and couldnt figure out where its coming from. I confirmed that not using the star allowed it to build. I wonder if something with gridsome/the graphql package hates that syntax 🤔 .

@pnadolny13 Yeah, the star is causing the generated graphql to break. I don't think the metadata is used to render the plugin pages, so wdyt about just overriding the value of `metadata` that is passed to make it null or a JSON string. We can then create an issue to correctly process the metadata if we want to render it in the future.

pnadolny13:
@edgarrmondragon thanks for that! I dont fully understand how that works but it fixed the problem and I trust you 😄 . I'm going to merge this.

_data/meltano/extractors/tap-mysql/singer-io.yml
```diff
@@ -9,9 +9,16 @@ keywords:
...
 maintenance_status: unknown
+metadata:
+  '*':
+    replication-method: FULL_TABLE
 name: tap-mysql
 namespace: tap_mysql
 pip_url: git+https://github.com/singer-io/tap-mysql.git
+prereq: |
+  ### Usage Notes
+  This tap is configured to use FULL_TABLE replication by default.
```
tayloramurphy, line 20:
```suggestion
  This tap is configured to use `FULL_TABLE` replication by default.
```

tayloramurphy (approved):

PR #1080, opened 2022-12-09
target-postgres has a default for port not obvious in meltano interac…

visch:
target-postgres has a default in target.py here https://github.com/MeltanoLabs/target-postgres/blob/main/target_postgres/target.py#L62 

This isn't reflected in the interactive window

image

I could add a `value` here as well? Maybe the automation needs this added @pnadolny13  ?

### DIFF
--- a/_data/meltano/loaders/target-postgres/meltanolabs.yml
+++ b/_data/meltano/loaders/target-postgres/meltanolabs.yml
@@ -23,7 +23,7 @@ settings:
   label: Host
   name: host
 - description: The port on which postgres is awaiting connection. Note if sqlalchemy_url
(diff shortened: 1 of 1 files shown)


pnadolny13:
@visch this is a good point, we've had this discussion in the past a few times. I went ahead and created https://github.com/meltano/hub/discussions/1081 to talk it through once and for all, and to have an artifact for the answer. Let me know what you think.

tayloramurphy (approved):

tayloramurphy:
Approving this change but good to have the discussion.

pnadolny13 (approved):

PR #1101, opened 2022-12-19
Add Airbyte extractors to the Hub

tayloramurphy:


pnadolny13 (commented):
@tayloramurphy should docker.png be dockerhub.png? I was planning to call it tap-dockerhub, I feel like tap-docker is not totally clear.

pnadolny13:
If we're going to add instructions to config the image name we need to keep track of any taps that dont match exactly tap-<x> to source-<x>, so I'll put them here for now and we can figure out how to handle them later:

- tap-apify -> source-apify-dataset

tayloramurphy:
@pnadolny13 looks like this revealed a Hub frontend bug around setting length https://deploy-preview-1101--meltano-hub.netlify.app/extractors/tap-alloydb

pnadolny13:
@tayloramurphy the GNews image looks to be for google news, which also tripped me up but the docs link out to https://gnews.io/. So that image will need to be replaced.

aaronsteers:
@tayloramurphy and @pnadolny13 - I noticed in our meeting that the config instructions use space as a delimiter, but I think we need to use dot (`'.'`) as delimiter.

aaronsteers:
@tayloramurphy and @pnadolny13 - The pages currently say "`tap-foobar from airbyte`". Can we change to something like `tap-foobar (airbyte variant)`?

tayloramurphy:
@aaronsteers I've made both of those changes! thanks for catching https://deploy-preview-1101--meltano-hub.netlify.app/extractors/tap-adjust

pnadolny13 (approved):
LGTM!

PR #1103, opened 2022-12-22
fix: Filter hidden variants from "Available variants" section

### LINKED ISSUE #897: Hide "hidden: true" variants from plugin info pages
Hidden plugins don't show up in search results or plugin index pages, but still show up in the "Other available variants" for non-hidden variants of the same target. We should hide these.

Example: https://hub.meltano.com/loaders/target-postgres
The now-defunct `meltano` variant is showing up in the transferwise variant page.

cjohnhanson:
Closes #897

### DIFF
--- a/src/templates/Plugins.vue
+++ b/src/templates/Plugins.vue
@@ -359,7 +359,7 @@ query Plugins($path: String!, $name: String!) {
       url
     }
   }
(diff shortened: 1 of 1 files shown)


tayloramurphy (approved):
Looks great!

pnadolny13 (approved):

PR #1128, opened 2023-01-16
Simplify config recommendation for Airbyte Connectors

tayloramurphy:
This removes the different instructions for how to configure Airbyte connectors as the nested config bug was fixed in 2.13.0.

Before: https://hub.meltano.com/extractors/tap-adjust

After: https://deploy-preview-1128--meltano-hub.netlify.app/extractors/tap-adjust


pnadolny13 (approved):
@tayloramurphy LGTM!

PR #1134, opened 2023-01-26
read metrics from new aws account s3 bucket

pnadolny13:
Related to https://github.com/meltano/squared/issues/247

- Squared is now writing to the new AWS account/bucket
- This makes the update to read from that bucket

Melissa created `Hub Metrics Bucket Read Access` in the 1Pass vault that is able to read the metrics files from the new S3 bucket. 

@rwfeather @cjohnhanson could one of you review this and also update the settings in this repo to use those new AWS credentials from 1Pass? I don't have access to do that.

cc @magreenbaum @tayloramurphy

### DIFF
--- a/utility_scripts/build/build_with_metrics.sh
+++ b/utility_scripts/build/build_with_metrics.sh
@@ -20,7 +20,7 @@ aws_access_key_id = $ENV_ACCESS_KEY_ID
 EOL
 
 # Grab files
-~/aws-cli/bin/aws s3 cp s3://prod-meltano-bucket-01/hub_metrics/variant_metrics.yml ./_data/variant_metrics.yml 
-~/aws-cli/bin/aws s3 cp s3://prod-meltano-bucket-01/hub_metrics/audit.yml ./_data/audit.yml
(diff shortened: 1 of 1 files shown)


pnadolny13:
@tayloramurphy do you have access to the settings of this repo? We just need to switch the credentials to be fully switched over to the new AWS account, then I can merge this.

magreenbaum:
@pnadolny13 Looks like just Netlify needs to be updated for these vars based on [this](https://github.com/meltano/hub/blob/ff09ec2279c7dc3bee05be116f61e47ebee680a1/utility_scripts/build/build_with_metrics.sh#L16):
`ENV_ACCESS_KEY_ID`
`ENV_SECRET_ACCESS_KEY`
`ENV_DEFAULT_REGION`

magreenbaum:
@pnadolny13 I've updated Netlify with the new creds.

magreenbaum (approved):
Not really sure why the netlify checks are still showing failed on the PR but it looked good when I reran in netlify. For me, this looks good outside of that but I'm not super familiar with netlify checks in GitHub.

pnadolny13:
@magreenbaum thanks, it looks like the preview is updated here as well so the check marks probably just arent getting updated. I think we're good to merge it.

Issue #1115, opened 2023-01-04
Make EDK preview utilities default


pnadolny13:
We still have the EDK based dbt utilities as "Preview" https://hub.meltano.com/utilities/dbt-snowflake. I think we had enough time to sort out any bugs.

- remove preview label from dbt utilities
- hide transformer dbt versions and file bundle
- make Airflow EDK utility the default
- hide the Airflow orchestrator/files bundle
- make Superset EDK utility the default

tayloramurphy:
@pnadolny13 there are a _ton_ of documentation updates that we will need to make to coincide with this change. This feels like a 3.0 level update since it'd such a change. @aaronsteers your thoughts?

aaronsteers:
@tayloramurphy - I think there's a path forward where we deprecate the older dbt implementation without removing support. That would allow us to update docs recommending the new dbt implementations as the primary/default choice, without necessarily needing a major version bump.

Wdyt?

pnadolny13:
@tayloramurphy yeah like AJ said, I was thinking that we could continue supporting as usual but update the hub to hide deprecated plugin types from searches and update the docs to recommend the utility syntax over orchestrators/transformers.

For docs I think a light lift for the first iteration would be to continue to refer to them as orchestrators/transformers and keep the docs the same except that example commands use `meltano add utility` instead of `meltano add transformer`. Even long term I dont hate calling them transformer utilities in docs even though theyre really all just plain utilities.

PR #1135, opened 2023-01-26
take edk utilities out of preview

pnadolny13:
Closes https://github.com/meltano/hub/issues/1115

- hides transformers, ochestrators, and associated files from search.
- removes the transformer and orchestrator options from the plugin drop down menu
- those pages still exist if you have a url and are still served via the API
- adds the edk repo url to the side bar in the plugin page below repo
- ~I drafted some changes to the Airflow plugin page to remove "preview" and the description. I figured once we get airflow in a good place I can update the other 6-7 utilities with the same language.~
  - ~do we still want a section to mention that its EDK based and what the EDK is? or does it not matter anymore?~
  - ~we need some consistency with our edk naming convention, some are named after meltano and some are named after the source package owner i.e. airflow is the apache variant, most dbt plugins are the dbt-labs variant, but dbt-duckdb is jwillis, and superset has both apache and meltano. I think if we use an edk then it should be the variant.~

Update: I added a section to every edk based plugin that explains what it is. I also adjusted how I hide those plugins from search so they still exists on plugin landings pages (i.e. transformers) but dont show up in search, this includes the associated files bundles.

@tayloramurphy what do you think?

Deploy preview:

- https://deploy-preview-1135--meltano-hub.netlify.app/

### DIFF
--- a/.github/workflows/build-api.yml
+++ b/.github/workflows/build-api.yml
@@ -63,6 +63,10 @@ jobs:
         --hub-url ${{ env.HUB_URL }}
         --api-url ${{ env.API_URL }}
 
+    - name: Test API Schema
+      run: >
+        poetry run python utility_scripts/api/api_schema_validate.py
+
     - name: Upload artifacts
       uses: actions/upload-artifact@v3.1.1
       with:
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ dist
 *.DS_Store
 .vscode
 .idea
+_hub_api*
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,7 +106,7 @@ Markdown pages are rendered using the `MarkdownDocs` template, using `/src/templ
 
 ### GraphQL query for markdown docs
 
-To better understand the data model for markdown data files, use the following query at the GraphQL endpoint (usually [http://localhost:8080/\_\_explore](http://localhost:8080/__explore)):
+To better understand the data model for markdown data files, use the following query at the GraphQL endpoint (usually [http://localhost:8080/\_\_explore](http://localhost:8080/___explore)):
 
 ```graphql
 query MarkdownDocs {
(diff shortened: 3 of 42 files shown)


tayloramurphy:
@pnadolny13 excited to see this. We need to pair this with a corresponding PR to the docs themselves as there are several resources that recommend the non-utility versions of these plugins. I chatted this morning with @sbalnojan and @aaronsteers about this and we're generally supportive of it, but there are some considerations for existing users that we may want to account for.

pnadolny13:
@tayloramurphy sounds good! Since the legacy pages still exist and are still served with the API I figured this wasnt blocked by the docs updates. We're only excluding these plugins from being discoverable on the hub.

pnadolny13:
> we need some consistency with our edk naming convention, some are named after meltano and some are named after the source package owner i.e. airflow is the apache variant, most dbt plugins are the dbt-labs variant, but dbt-duckdb is jwillis, and superset has both apache and meltano. I think if we use an edk then it should be the variant.

@tayloramurphy related to variant names and ^^ I still think Superset is still a weird case. We have apache for the direct pip install and meltano for the extension version, both would be apache if we went by the source code naming. I also think theres a chance that we will get multiple EDK variants that have different features, what do you think about doing `<source code org name>-<edk extension org name>` (or a similar combo of the 2 names) as the variant so we'd have:
- `apache` and `apache-meltano` for superset
- dbt-snowflake -> `dbt-labs-meltano`
- dbt-duckdb -> `jwills-meltano`

pnadolny13:
@tayloramurphy I made the updates that we talked about. Let me know what you think.

tayloramurphy:
@pnadolny13 what do you think about hiding the related files (dbt, airflow)? That would bring it back down to just datahub, docker, docker compose, gitlab ci, and great expectations.

pnadolny13:
@tayloramurphy hiding from search and the files plugin page? Right now it should already be hidden in search but I can switch to remove them from the plugin page too. I think that makes sense and might be less confusing for users to not see those at all.

tayloramurphy:
@pnadolny13 I just pushed up a change to hide the dbt files.

tayloramurphy:
Docs PR is merged. We can merge this on 2/3 it looks like.

PR #1146, opened 2023-02-01
add api schema tests back in

pnadolny13:
@tayloramurphy as part of adding `hidden` flags to files bundles [we got a test failure](https://github.com/meltano/hub/pull/1135#issuecomment-1412365864) because the schemas dont accept those flags. This also raised to me the fact that we stopped testing the schema of the generated API which is concerning because what we allow in the plugin definitions and in the API are not the same e.g. usage, next_steps, etc. Luckily nothing is violating the schema though 😅 .

I can make `hidden` an acceptable attribute in the plugin definition but I'm not sure if we want it to be exposed as part of the API, like it is today with extractors/loaders. I'd lean towards no because we're mostly using the flag to hide or show in the hub but I know the hidden attribute also affects what shows up in the deprecated meltano UI.

### DIFF
--- a/.github/workflows/build-api.yml
+++ b/.github/workflows/build-api.yml
@@ -63,6 +63,10 @@ jobs:
         --hub-url ${{ env.HUB_URL }}
         --api-url ${{ env.API_URL }}
 
+    - name: Test API
+      run: >
+        poetry run python utility_scripts/api/api_schema_validate.py
+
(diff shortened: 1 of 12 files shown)


tayloramurphy:
I think it makes sense to leave it out. It's primarily for the UI like you said.

tayloramurphy (approved):

PR #1155, opened 2023-02-07
fix command docs, remove args

pnadolny13:
We currently show a [command section](https://hub.meltano.com/utilities/airflow#commands) on the utility pages that says you can run `meltano invoke airflow:create-admin [args...]` but I think this is a typo. You either have to not include the colon to run with args or no args are allowed. I'm not sure what way we want to recommend in the docs but this PR removes the reference to args.

### DIFF
--- a/src/components/PluginCommandsSection.vue
+++ b/src/components/PluginCommandsSection.vue
@@ -3,7 +3,9 @@
     <p class="text-3xl py-4" id="commands">Commands</p>
     <span
       >The {{ name }} {{ plugin_type }} supports the following commands that can be used with
-      <pre><code>meltano invoke</code></pre>
+      <a href="https://docs.meltano.com/reference/command-line-interface#invoke">
+        <pre><code>meltano invoke</code></pre>
+      </a>
       :</span
     >
     <div v-for="(command, key, index) in commands" v-bind:key="index">


pnadolny13:
@tayloramurphy interesting, it looks like my understand was wrong 😞 . Thanks for that docs link. I thought that users could either use the colon to execute meltano commands with no support for arguments or if they did want to use arguments then they call the executable directly by excluding the colon. 

Testing a couple commands to clarify my understanding:

- `meltano invoke dbt-snowflake:seed --help` ✅ 
- `meltano invoke dbt-snowflake seed --help` ✅ 
- `meltano run dbt-snowflake:seed` ✅ 
- `meltano run dbt-snowflake:seed --help` ❌  returns the generic help text for the `meltano run` command same as if you ran `meltano run --help`

I can close this PR unless you still want me to make those other changes.

tayloramurphy:
@pnadolny13 that makes sense though - it only works on invoke and run doesn't accept additional args. 

@pnadolny13 I like this PR - just update the `meltano invoke` reference at the start of the section to link to the docs and I think we're good.

pnadolny13:
@tayloramurphy I updated it to include the link!

tayloramurphy (approved):
Looks good!

src/components/PluginCommandsSection.vue
```diff
@@ -19,7 +21,7 @@
       <p>{{ command.description }}</p>
       <pre
         class="prose language-bash rounded-md"
-      ><code>meltano invoke {{ name }}:{{ command.name }} [args...]</code></pre>
+      ><code>meltano invoke {{ name }}:{{ command.name }}</code></pre>
```
aaronsteers, line 24: This is correct as written, and I think noting `args...` as an optional input is helpful so users can discover this feature.

Should we revert this part back?

```suggestion
      ><code>meltano invoke {{ name }}:{{ command.name }} [args...]</code></pre>
```

As noted in the PR thread, the added args are freely supported in `invoke` - just not in `run`.

aaronsteers (changes requested):

PR #1169, opened 2023-02-14
fix superset docs and pin conflicting dependency

pnadolny13:
Related to https://github.com/meltano/superset-ext/pull/8

- The docs were outdated for the EDK variant using superset_ext which isnt what the extension expects anymore
- Theres an issue in the [superset dependencies](https://github.com/apache/superset/issues/22571) that causes an error unless we explicitly pin a version of `cryptography` also see [MDS in a box where the same thing was done](https://github.com/matsonj/nba-monte-carlo/issues/63). I updated for both of these variants
- Set the EDK variant to default

### DIFF
--- a/_data/default_variants.yml
+++ b/_data/default_variants.yml
@@ -598,4 +598,4 @@ utilities:
   notebook: matatika
   postgres: postgres
   sqlfluff: sqlfluff
-  superset: apache
+  superset: meltano
--- a/_data/meltano/utilities/superset/apache.yml
+++ b/_data/meltano/utilities/superset/apache.yml
@@ -67,7 +67,7 @@ next_steps: |
      utilities:
      - name: superset
        variant: apache
-       pip_url: apache-superset==1.5.0 snowflake-sqlalchemy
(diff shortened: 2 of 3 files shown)


cwegener:
@pnadolny13 The `cryptography` version issue fix is planned to be included in the 2.0.2 release.

pnadolny13:
> @pnadolny13 The `cryptography` version issue fix is planned to be included in the 2.0.2 release.

@cwegener thanks for that! I went looking and found that the release should be out next week and you linked to https://github.com/apache/superset/pull/22744 that fixes the problem.

I'd vote to merge this ASAP, then open an issue to bump to 2.0.2 + remove the cryptography pin after the release next week.

tayloramurphy (approved):
I'm still not a fan of having this as the Meltano variant, but it's not critical we solve for that problem right now. Switching to the EDK as default makes sense to me.

pnadolny13:
> I'm still not a fan of having this as the Meltano variant, but it's not critical we solve for that problem right now. Switching to the EDK as default makes sense to me.

@tayloramurphy yeah thats fair, cross linking back to our previous discussion on this https://github.com/meltano/hub/pull/1135#issuecomment-1408992223. I'll open an issue so it doesn't get lost.

PR #1186, opened 2023-03-07
Add `matatika` utility

ReubenFrankel:
## Checklist

- [x] Add/update the file in the appropriate folder (`/taps` or `/targets`). The name of the file should match the name of the tap. If there is already one, add a descriptor to the name such as `-search`.
- [x] Add/update the PNG logo image in `/assets/logos/<taps or targets>`. The image name must match the YAML file name.
- [x] Tag `@tayloramurphy` or `@pnadolny13` to flag it for review. Or post to the [#hub](https://meltano.slack.com/archives/C01UGBSJNG5) channel on Meltano slack.

## Reviewer Checklist

- [ ] Validate file against JSON Schema. https://www.jsonschema.net/home is an option.
- [ ] Build website locally and validate everything works.

PR #1189, opened 2023-03-08
Fix Matatika utility malformed `pip_url` / minor content updates

ReubenFrankel:
Fix following merge of #1186. In the future, it would be nice to have a CI check here that attempts to `pip install` from the supplied `pip_url` to catch potential `meltano install` issues.


ReubenFrankel:
cc @pnadolny13

aaronsteers (approved):

Issue #1190, opened 2023-03-08
"If you run into any issues, learn how to get help." doesn't make sense


ReubenFrankel:
### Problem

> If you run into any issues, learn how to get help.

This text shows for any plugin page at the end of the `Next steps` section, which doesn't make any sense by itself.

https://github.com/meltano/hub/blob/f19c9889923fcd6c42267c6859e53614ea36fe1c/src/templates/Plugins.vue#L277-L278

### Potential solution
As a contributor, I would like to be able to specify `help` in my plugin `.yml` file and have that rendered as Markdown instead.

```yml
help: |
  To get help, join our [Slack community](<link>) and make a post in `#troubleshooting` 
```

tayloramurphy:
@ReubenFrankel I believe that was originally supposed to link to the section at the bottom of the page. I've got a PR #1193 to fix that. 

As for the `help` section, we do have the `usage` key which can be used for that somewhat. For now I'm going to say no to adding a new section, but I'm definitely open to PRs to update the help section for all plugins https://github.com/meltano/hub/blob/main/src/components/PluginHelpSection.vue

ReubenFrankel:
@tayloramurphy For the `matatika` utility, I ended up linking to our Slack in the [`next_steps` section](, since `usage` to me (at the time

ReubenFrankel:
@tayloramurphy For the `matatika` utility, I ended up [linking to our Slack in the `next_steps` section](https://github.com/meltano/hub/blob/f19c9889923fcd6c42267c6859e53614ea36fe1c/_data/meltano/utilities/matatika/matatika.yml#L37-L52), but I can see how it might fit better under `usage` now you mention it... Ideally, I think it should live under the `Looking for help?` section, because that's where I would to if I was having a problem (as a user of the plugin). Is that what you were suggesting when you said

> but I'm definitely open to PRs to update the help section for all plugins

(i.e. show a custom Matatika help section for plugins specifying `matatika` in `keywords`, [like how `airbyte_protocol` is handled currently](https://github.com/meltano/hub/blob/f19c9889923fcd6c42267c6859e53614ea36fe1c/src/components/PluginHelpSection.vue#L11-L24))?

tayloramurphy:
@ReubenFrankel sorry for the confusion. 

My point about updating the Help section for all plugins was that if there was something better we could say to point Meltano users to more helpful resources, I'm all for that. But that editing that file would update it for all plugins. 

[...]

PR #1193, opened 2023-03-08
add link to anchor

tayloramurphy:
Closes https://github.com/meltano/hub/issues/1190


pnadolny13 (approved):
@tayloramurphy I just pushed a `yarn format:write` change that should fix the failing lint test. Other than that looks good to me!

Issue #1095, opened 2022-12-16
Automate metadata refresh for SDK plugins


pnadolny13:
What:
- We should have an automated process that checks the `--about` output of SDK based plugins and refreshes the hub definition. 

Why:
- Since we prefer SDK based variants and know that they have better stability and better levels of maintenance, we should also be sure that the metadata is always accurate on the hub. We rely on user to update these today which allows some to get out of sync and unchecked.
- Bad settings and metadata can cause the plugin not to work in meltano, so I think this is an important thing to be maintaining.
- This also encourages plugin developer to put their documentation in that plugin repo itself as the SSOT and let the hub scrape it.

How:
- It can run daily/weekly to guarantee the metadata we publish about SDK based variants are good. I created this script https://github.com/pnadolny13/hub-utils that does the core of this translation already.

Challenges:
- I found some functioning taps that fail during help/about so those would have to be skipped
- python version limitations would have to be considered. Some of our CI tests fail to extract about metadata because the python version of the test container isnt supported by the tap.

aaronsteers:
I love this idea. I think it's save to skip those plugins which fail in `--about` for some reason. It could also be helpful to have a list of those and even open an issue on all repos for which `--about` cannot complete.

In an early version of the SDK, some taps or targets could not complete execution of `--about` unless valid config was provided, but in more recent version this should not be an issue.

Another benefit of automating this, is that we can give developers the option to link their README.md back to the Hub "for the latest settings documentation". This could complement or replace their own `settings` list which periodically need to be updated manually via markdown edits.

pnadolny13:
I spent a little time thinking about how this process could work for updating SDK variants and since its relatively similar for Airbyte variants too I want to consider them in this also. I proved out that I can:

1. iterate all existing hub plugins and find a list thats SDK based
2. install the executable via pipx and run --about
3. translate the --about output to the hub definition spec
4. merge the existing definition and the newly exported --about output i.e. updating the existing definition in the hub repo

### Challenges:

1. The install/--about process is pretty slow since we already have ~130 SDK variants
2. Translating --about to the hub definition spec can have edge cases that I dont know about yet. It also requires me to be opinionated about things like sort ordering so that the diffs are consistent between runs. For example I chose to sort the settings list of dicts in alpha order using the name key. 
3. Failures are somewhat frequent ~25/130 failed on my first iteration.
4. Some failures are related to python versions
5. Some failures are due to required configs or credentials for discovery.
6. Airbyte variants require docker 

[...]

PR #1208, opened 2023-03-24
Metadata Refresh Github Action - Part 1

pnadolny13:
Related to https://github.com/meltano/hub/issues/1095

This is part 1 of a 2 step process to automate refreshing hub metadata for SDK based and Airbyte variants. The first step is to automate extracting the `--about` output for SDK variants and SPEC output for airbyte variants, which this PR does.

See https://github.com/pnadolny13/hub/actions/runs/4513484433 as an example of this running, it shows it failing at the top level summary but thats because some of the variants are broken and is expected. In 28 mins it extracted metadata from 406 variants across SDK and airbyte images 🚀 .

- This relies heavily on my [hub_utils](https://github.com/pnadolny13/hub-utils) CLI. I'll admit that it needs some cleanup work now that the workflow is becoming more clear but it does what it needs to do for this part.
- Steps:

### 1.) get_variants_list
This calls the hub_utils CLI which iterates the hub directory structure (i.e. checking the keyword list for `meltano_sdk` or `airbyte_protocol`) and returns a list of variants that we should extract metadata for. Its split up into 3 buckets right now: SDK extractors, SDK loaders, and airbyte extractors. Airbyte and SDK need to be extracted differently so theyre separate and because Github limits a matrix to 256 jobs which we were going over. 

The output of the CLI is set as job output so we can reference it in downstream jobs

### 2.) metadata_extract_sdk_ex and metadata_extract_sdk_loaders
These are basically the same minus the fact that they use 2 different matrix inputs (i.e. extractors and loaders), eventually we'd hit the 256 limit so I went ahead and separated them to start. For each variant in the matrix we install and run the hub_utils CLI to attempt to extract the `--about` metadata. It starts with python 3.10 then if that fails it tries 3.9 then 3.8. If the first python version succeeds then the rest are skipped.

The [hub_utils CLI command](https://github.com/pnadolny13/hub-utils/pull/10) runs the `--about` command then uploads the output to S3 using a file name based on the hash of the contents and the current date. This allows us to only upload when the data changes and to also have the date so we know which hash is the most recent.

Its possible that these two jobs could be combined into a fancier double matrix job since the steps are duplicated or we could use a python matrix instead explicitly catching failures...but thats for a future iteration.

### 3.) metadata_extract_airbyte_p1

This step also uses a matrix given the output from `get_variants_list`. This one uses the container feature to open a shell inside the airbyte container we want to extract data from. Once inside we can use the `$AIRBYTE_ENTRYPOINT` environment variable thats required in every connector as an entrypoint in order to get run `$AIRBYTE_ENTRYPOINT spec`. That command outputs the message we want to save. It also outputs other log json messages and some connectors output pure application logs too so I grep for the single SPEC json record we're looking for. Then I couldnt figure out how to upload that output to S3 from within the container because theres a bunch of different base images used so python was unreliable and installing aws cli sounded difficult. My solution is to set it as a job artifact that the next part can access.

In a future iteration I might try configuring python and the hub_utils CLI in each container to avoid the artifacts and the 2 different jobs.

### 4.) metadata_extract_airbyte_p2

Similarly this step also uses a matrix given the output from `get_variants_list`. It basically just pulls down the part 1 job artifact for the particular variant then uses the hub_utils CLI to upload it to S3 using the special hashed file name logic. This is primarily because I couldnt upload to S3 from part 1.

# Secret Requirements

I've been using a test S3 bucket from Squared. For this to work we need to configure AWS credentials in github actions.


pnadolny13:
Interesting to note also that I was able to do a full run yesterday and then again today and I saw ~5 plugins that had updated metadata. I'm very curious to see how far the hub has drifted, part 2 should answer that question.

tayloramurphy (approved):
Amazing work Pat! I'm not one to actually review the actions yaml - but if it works then you're good.

PR #1220, opened 2023-03-31
feat: provide matching default db path for `target-duckdb` and `dbt-duckdb`

aaronsteers:
This would use the `output` folder as a default location for the generated duck DB database file. I've applied the same default path for both `target-duckdb` and `dbt-duckdb`, so if users do not want to override a path, they will by default get both working together.

Specifically: `warehouse.duckdb` in `${MELTANO_PROJECT_ROOT}/output`

The reason for this change is so that we quickly swap in `target-duckdb` for speedrun invocations like this one in slack: https://meltano.slack.com/archives/C04632W0HT2/p1680301581525659

I don't believe this is a breaking change since previously both plugins would simply fail if paths were not provided.

### DIFF
--- a/_data/meltano/utilities/dbt-duckdb/jwills.yml
+++ b/_data/meltano/utilities/dbt-duckdb/jwills.yml
@@ -97,6 +97,7 @@ settings:
   kind: string
   label: Path
   name: path
+  value: "${MELTANO_PROJECT_ROOT}/output/warehouse.duckdb"
 - description: Specify the schema to write into.
   kind: string
   label: Schema
--- a/_data/meltano/utilities/dbt-duckdb/jwills.yml
+++ b/_data/meltano/utilities/dbt-duckdb/jwills.yml
@@ -97,7 +97,7 @@ settings:
   kind: string
   label: Path
   name: path
-  value: "${MELTANO_PROJECT_ROOT}/output/warehouse.duckdb"
+  value: ${MELTANO_PROJECT_ROOT}/output/warehouse.duckdb
 - description: Specify the schema to write into.
   kind: string
   label: Schema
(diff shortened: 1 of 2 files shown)


_data/meltano/utilities/dbt-duckdb/jwills.yml
```diff
@@ -97,6 +97,7 @@ settings:
   kind: string
   label: Path
   name: path
+  value: "${MELTANO_PROJECT_ROOT}/output/warehouse.duckdb"
```
aaronsteers, line 100:
```suggestion
  value: ${MELTANO_PROJECT_ROOT}/output/warehouse.duckdb
```

edgarrmondragon:
I think target-duckdb still would require either setting `default_target_schema` or `schema_mapping`

aaronsteers:
> default_target_schema

Correct! Thanks for calling this out... 🤔

aaronsteers:
> > I think target-duckdb still would require either setting `default_target_schema` or `schema_mapping`
> 
> Correct! Thanks for calling this out... 🤔

@edgarrmondragon - I think Meltano might provide this default if invoked with Meltano. I'll test though before I move out of draft status.

aaronsteers:
Note: the first version of this PR only updated the target. Now the PR includes symmetrical update also for `dbt-duckdb`.

aaronsteers:
> > > I think target-duckdb still would require either setting `default_target_schema` or `schema_mapping`
> > 
> > 
> > Correct! Thanks for calling this out... 🤔
> 
> @edgarrmondragon - I think Meltano might provide this default if invoked with Meltano. I'll test though before I move out of draft status.

Confirmed. We should be okay here. On a local project I added only `filepath` config and it ran with no issues.

aaronsteers:
I think this is ready to go. I don't have `dbt-duckdb` confirmed as zero-config, but at least paths are aligned by default. Some additional steps are probably needed on the dbt side to get a working transform running, but for now I think this makes a nice iteration forward.

PR #1225, opened 2023-04-04
assume role for hub metadata extract

pnadolny13:


edgarrmondragon (commented):
@pnadolny13 Since these jobs can only triggered by `workflow_dispatch`, I think we can make the environment an input:

```yaml
on:
  workflow_dispatch:
    environment:
      bump:
        description: "Hub Environment"
        required: true
        type: choice
        options:
        - production
        - preview
...

  metadata_extract_sdk_ex:
    name: SDK Extractors - Metadata Extract
    needs: get_variants_list
    runs-on: ubuntu-latest
    environment: ${{ github.event.input.environment }}
```

pnadolny13:
> @pnadolny13 Since these jobs can only triggered by `workflow_dispatch`, I think we can make the environment an input:
> 
> ```yaml
> on:
>   workflow_dispatch:
>     inputs:
>       environment:
>         description: "Hub Environment"
>         required: true
>         type: choice
>         options:
>         - production
>         - preview
> ...
> 
>   metadata_extract_sdk_ex:
>     name: SDK Extractors - Metadata Extract
>     needs: get_variants_list
>     runs-on: ubuntu-latest
>     environment: ${{ github.event.input.environment }}
> ```

@edgarrmondragon to be honest I'm not quite sure how to interpret what `${{ github.event_name == 'push' && github.ref == 'refs/heads/main' && 'production' || 'preview' }}` is doing. I agree with the idea to add input if its only run manually but eventually I'll want this on a schedule so I'm not sure if that will require a change again.

Also it looks like the secrets/role assuming is working now. I've got data in the staging environment!

edgarrmondragon:
> @edgarrmondragon to be honest I'm not quite sure how to interpret what `${{ github.event_name == 'push' && github.ref == 'refs/heads/main' && 'production' || 'preview' }}` is doing. I agree with the idea to add input if its only run manually but eventually I'll want this on a schedule so I'm not sure if that will require a change again.
> 
> Also it looks like the secrets/role assuming is working now. I've got data in the staging environment!

The current setup will only ever deploy to staging/preview since . You can think of what it's currently doing as roughly equivalent to

```sql
SELECT
  CASE
    WHEN event_type = 'push' AND target_branch = 'main' THEN 'production'
    ELSE 'preview'
  END AS enviroment
```

Let me send you a PR against this branch

tayloramurphy (approved):
Approving so you can merge when ready.

PR #1230, opened 2023-04-05
fix: Adjust environment conditional for metadata extract action

magreenbaum:
This conditional is not actually setting 'production' as the environment. 

Ran a few tests on another branch. Basically getting a failed when setting credentials error where it's not using an env at all. Switching to removing the conditional so it uses the input environment and when selecting `production` now I am at least getting a job failed status where it doesn't attempt to run the action at all since the production env is protected by a branch rule for main only deployments. Selecting preview works as expected since preview can run on any branch.

### DIFF
--- a/.github/workflows/metadata-extract.yml
+++ b/.github/workflows/metadata-extract.yml
@@ -62,7 +62,7 @@ jobs:
     name: SDK Extractors - Metadata Extract
     needs: get_variants_list
     runs-on: ubuntu-latest
-    environment: ${{ github.event.inputs.environment == 'production' || github.event_name == 'schedule' && 'production' || 'preview' }}
+    environment: ${{ github.event.inputs.environment || github.event_name == 'schedule' && 'production' || 'preview' }}
     env:
       AWS_S3_BUCKET: "${{secrets.HUB_METADATA_S3_BUCKET }}"
     strategy:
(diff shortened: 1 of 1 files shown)


pnadolny13:
@magreenbaum if it works it works! I didnt think about the main branch protection.

Personally I'm confused whats happening though, previous Edgar said the logic was equivalent to:

```
if github.event.inputs.environment == 'production' or github.event_name == 'schedule':
    environment = 'production'
else:
    environment = 'preview'
```

So I'd read the new logic as this, right?:

```
if github.event.inputs.environment  or github.event_name == 'schedule':
    environment = 'production'
else:
    environment = 'preview'
```

To me that means setting the input environment to preview in the dropdown will return production. Am I misunderstanding the logic?

magreenbaum:
> @magreenbaum if it works it works! I didnt think about the main branch protection.
> 
> Personally I'm confused whats happening though, previous Edgar said the logic was equivalent to:
> 
> ```
> if github.event.inputs.environment == 'production' or github.event_name == 'schedule':
>     environment = 'production'
> else:
>     environment = 'preview'
> ```
> 
> So I'd read the new logic as this, right?:
> 
> ```
> if github.event.inputs.environment  or github.event_name == 'schedule':
>     environment = 'production'
> else:
>     environment = 'preview'
> ```
> 
> To me that means setting the input environment to preview in the dropdown will return production. Am I misunderstanding the logic?

My understanding of the first bit `github.event.inputs.environment` is whatever you choose for the input is what will be set here. Otherwise `github.event_name == 'schedule' && 'production'` will be production, and else is preview.

pnadolny13 (approved):
@magreenbaum cool lets give it a shot!

PR #1231, opened 2023-04-06
set z3z1ma as default tap-mongodb

pnadolny13:
@z3z1ma what do we think about making your tap the default? Its SDK based and sounds like from this thread https://meltano.slack.com/archives/C04632W0HT2/p1674951678271239 its a good generic implementation for schema-less sources. I was referencing it for a new tap-dynamodb and noticed this isnt default yet. Any reason it isn't ready to be default? Like known features that [transferwise](https://github.com/transferwise/pipelinewise-tap-mongodb) supports that this variant doesnt.


tayloramurphy:
@pnadolny13 is there a slack thread where this is being discussed?

pnadolny13:
@tayloramurphy I just started a thread in https://meltano.slack.com/archives/C01UGBSJNG5/p1681482241550669

PR #1235, opened 2023-04-10
fix: remove state and catalog capabilities from tap-stackoverflow-sampledata

BuzzCutNorman:
When I initially submitted the tap to the hub I put `state` and `catalog` as capabilities but these should have not been added.  The tap was not written to do either.  This will close a tap issue that a I opened a while back.

https://github.com/BuzzCutNorman/tap-stackoverflow-sampledata/issues/4

## Checklist

## Reviewer Checklist


BuzzCutNorman:
@pnadolny13 when you get a chance, please flag this PR for review. 🙏

_data/meltano/extractors/tap-stackoverflow-sampledata/buzzcutnorman.yml
```diff
@@ -1,6 +1,4 @@
 capabilities:
-- state
-- catalog
 - discover
```
pnadolny13, line 2: @BuzzCutNorman its possible that this is related to a few issues in meltano like https://github.com/meltano/meltano/issues/6919. It sounds like from https://github.com/BuzzCutNorman/tap-stackoverflow-sampledata/issues/4 meltano is passing in an out dated catalog after you updated the input files. 

Deactivating these capabilities seems reasonable especially for a sample data tap. I'm now wondering what the purpose of discover is if it doesnt accept a catalog. I think you already know this but wanted to add the context in case its helpful, raw Singer workflow is usually to discover the catalog `tap-x --discover > catalog.json` then pass that output in as input to the tap `tap-x --catalog catalog.json`. I guess it still has that capability though, so we can leave it here.

pnadolny13 (approved):

_data/meltano/extractors/tap-stackoverflow-sampledata/buzzcutnorman.yml
BuzzCutNorman: @pnadolny13 Yes, you are correct the scenario I am running into fits the issue sited and would be resolved with the use of `--refresh`.  In the context of my desired workflow, I would be running the tap constantly with it.  The workflow I had in mind for the tap is that you can change what files are present in the folder between runs.  A static catalog file breaks that workflow.  I see `--discover` as a way to view what files the tap sees in the designated directory before a run.  The `--catalog` is used by default when I use the tap with `meltano invoke` or `meltano run` which is good in most cases just not this instance in which the source files present in the directory could change between runs.  I hope this is the information you are wanting, and it answers your questions if it doesn't, please let me know.

pnadolny13: @BuzzCutNorman that makes total sense, thanks for explaining. I'm good with removing that capability to inform Meltano that catalogs shouldnt be used.

PR #1256, opened 2023-04-14
Feat: Most popular extractors

javajunky:
Simple accordion of most popular extractors.


pnadolny13:
@javajunky no rush but what do you think this would take to be ready to merge?

I think we talked about, and I'd also prefer, the option to manually set a few of the most popular then let the usage dictate the rest. Sometime the data gets skewed due to big spikes likely from testing or something so keeping the set of the ones that we know are the top always in there would be great.

tayloramurphy (approved):
I'm going to merge this. It looks good and we can iterate on the query.

Issue #1255, opened 2023-04-14
Add SDK tap-criteo variant


pnadolny13:
https://github.com/edgarrmondragon/tap-criteo

PR #1258, opened 2023-04-17
Adds extractors tap-criteo (edgarrmondragon)

pnadolny13:
Closes https://github.com/meltano/hub/issues/1255.

Mentioned in Slack https://meltano.slack.com/archives/CMN8HELB0/p1681480173524249

@edgarrmondragon let me know if this should be listed on the hub and if this definition looks right. The other variants dont look too good so if you think yours is in good shape we might also want to consider making it the default.

### DIFF
--- a/_data/meltano/extractors/tap-criteo/edgarrmondragon.yml
+++ b/_data/meltano/extractors/tap-criteo/edgarrmondragon.yml
@@ -1,0 +1,87 @@
+capabilities:
+- about
+- catalog
+- discover
+- schema-flattening
+- state
+- stream-maps
+description: Advertising Platform
+domain_url: https://www.criteo.com/
+executable: tap-criteo
+keywords:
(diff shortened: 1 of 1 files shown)


tayloramurphy (approved):
I vote for making it the default!

_data/meltano/extractors/tap-criteo/edgarrmondragon.yml
```diff
@@ -0,0 +1,87 @@
+capabilities:
+- about
+- catalog
+- discover
+- schema-flattening
+- state
```
edgarrmondragon, line 6:
```suggestion
```

edgarrmondragon (commented):
@pnadolny13 I'm not sure if my variant has good coverage at this point (I have not updated functionality in almost 2 years 😬), but I can confirm that at least what it covers works well.

PR #1264, opened 2023-04-18
docs: incremental replication to tap-postgres

pnadolny13:
I was setting this up locally and a hint on the hub would have saved me time. This info is in the docs but I didnt find it easily accessible in one place and the metadata required for this was slightly different between variants.

### DIFF
--- a/_data/meltano/extractors/tap-postgres/meltanolabs.yml
+++ b/_data/meltano/extractors/tap-postgres/meltanolabs.yml
@@ -28,8 +28,8 @@ settings:
   kind: integer
   label: Flattening Max Depth
   name: flattening_max_depth
-- description: Example postgresql://postgres:postgres@localhost:5432/postgres
-  kind: string
+- description: Example postgresql://[username]:[password]@localhost:5432/[db_name]
+  kind: password
   label: Sqlalchemy URL
   name: sqlalchemy_url
 - description: User-defined config values to be used within map expressions.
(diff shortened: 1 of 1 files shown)


_data/meltano/extractors/tap-postgres/meltanolabs.yml
```diff
@@ -28,8 +28,8 @@ settings:
   kind: integer
   label: Flattening Max Depth
   name: flattening_max_depth
-- description: Example postgresql://postgres:postgres@localhost:5432/postgres
-  kind: string
+- description: Example postgresql://[username]:[password]@localhost:5432/[db_name]
+  kind: password
```
tayloramurphy, line 32: @pnadolny13 I actually don't know if this should be a password type. most folks like to see the full URL and if they need to pass a password in it can be via an environment variable reference right?

pnadolny13: @tayloramurphy good call out - I saw https://github.com/MeltanoLabs/tap-postgres/issues/5 also.

Personally I thought it seemed out of character for Meltano to put my password in my git back meltano.yml file vs in my .env by default. It seems risky to do that without a big warning for the user to remove it from your yaml file manually before checking it into git.

tayloramurphy: @pnadolny13 that's a fair point. I'd be okay with having it as a password so that in interactive mode we're not at risk of storing anything sensitive. Users can still override it if need be. 

I like the idea to break things out as well.

tayloramurphy (approved):

PR #1283, opened 2023-04-28
meltanolabs target-postgres to default

pnadolny13:
- target-postgres from transferwise-> meltanolabs

Is it ready to be default?

### DIFF
--- a/_data/default_variants.yml
+++ b/_data/default_variants.yml
@@ -556,7 +556,7 @@ loaders:
   target-oracle: radbrt
   target-pardot: anelendata
   target-parquet: estrategiahq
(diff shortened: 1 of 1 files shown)


tayloramurphy (commented):
@visch thoughts?

visch:
> @visch thoughts?

We could go for it, but there's still open known issues. Tap postgres got some love this week and is looking pretty good.

Target postgres is going to get some work done from me this coming week so it will be in a better spot, but honestly it just comes down to when do we want to ship it as the more folks that use it the more help we will get. 

Techy me wants to delay until everything is right

pnadolny13:
This is already completed

Issue #989, opened 2022-11-01
bug: tap-recruitee install fails


pnadolny13:
https://hub.meltano.com/extractors/tap-recruitee

```
Traceback (most recent call last):
  File "/Users/pnadolny/Documents/Git/GitHub/pnadolny/hub-utils/hub_utils/test_meltano_project/.meltano/extractors/tap-recruitee/venv/bin/tap-recruitee", line 5, in <module>
    from tap_recruitee import main
  File "/Users/pnadolny/Documents/Git/GitHub/pnadolny/hub-utils/hub_utils/test_meltano_project/.meltano/extractors/tap-recruitee/venv/lib/python3.8/site-packages/tap_recruitee/__init__.py", line 1, in <module>
    import singer
ModuleNotFoundError: No module named 'singer'
```

edgarrmondragon:
@pnadolny13 The `setup.py` file [is doing some unorthodox stuff](https://github.com/Tahasadiki/tap-recruitee/blob/3fbbf7b2baf37240223bad1e6644bc05b329dbbf/setup.py) by suggesting it be edited locally when installing the tap (as non-editable) so dependencies are actually installed along-side...

A workaround on our side might be to update our `pip_url` to include those packages:

```yaml
pip_url: >
  git+https://github.com/Tahasadiki/tap-recruitee.git
  attrs==18.1.0
  backoff==1.3.2
  python-dateutil==2.7.3
  requests==2.20.0
  singer-python==5.3.3
```

PR #1313, opened 2023-05-11
Recruitee refresh

pnadolny13:
Closes https://github.com/meltano/hub/issues/989

- fixes the bug by pinning additional dependencies explicitly as suggested by @edgarrmondragon. I'm now able to install and run `--help` successfully.
- add @rawwar's new SDK variant - It looks like its not completely done yet but its set as `development` status and will let others know that theres something thats work in progress. Also it opens the option for others to join forces with you if they were thinking about rebuilding this tap also.

### DIFF
--- a/_data/meltano/extractors/tap-recruitee/rawwar.yml
+++ b/_data/meltano/extractors/tap-recruitee/rawwar.yml
@@ -1,0 +1,63 @@
+capabilities:
+- about
+- catalog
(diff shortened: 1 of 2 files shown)


rawwar:
I will be completing this by next Wednesday. Will notify here once I'm done.

tayloramurphy (approved):

PR #1317, opened 2023-05-12
Add GtheSheep Netsuite variant and Talon One tap

GtheSheep:
## Checklist

- [x] Add/update the file in the appropriate folder (`/taps` or `/targets`). The name of the file should match the name of the tap. If there is already one, add a descriptor to the name such as `-search`.
- [x] Add/update the PNG logo image in `/assets/logos/<taps or targets>`. The image name must match the YAML file name.

## Reviewer Checklist

@pnadolny13


pnadolny13 (approved):
@GtheSheep thanks for the PR, this looks great!

One question - I saw [a slack thread](https://meltano.slack.com/archives/C04632W0HT2/p1682110344585129?thread_ts=1682108846.579989&cid=C04632W0HT2) you were in about the difference between REST/SOAP taps. Is there anything thats worth noting in the `usage` section around things that are supported or not supported by this tap since its REST vs SOAP? Or is it just an implementation detail that a user wouldnt care about? I'm not personally familiar with Netsuite.

I'm imagining once theres both a SDK tap using REST and one for SOAP then users might be confused which one is which and why they would chose one over the other.

GtheSheep:
Hey @pnadolny13 - yeah good call, there is! I think I marked the REST tap as Beta here too because the REST API isn't as complete as the SOAP API and some of the endpoints (though, not the ones in my tap) are still in Beta, so it's still a bit of a work in progress, contains fewer transaction types, etc, is this the kind of detail you're referring to? I'll add if so!

pnadolny13:
>is this the kind of detail you're referring to? I'll add if so!

@GtheSheep yes! Any detail that would help future users understand when to use your tap over others and if there are any obvious pros/cons to consider would be super helpful.

Issue #1327, opened 2023-05-18
Update maintainer label for Meltano-maintained connectors


tayloramurphy:
https://hub.meltano.com/extractors/tap-linkedin-ads is an example where we are maintaining it as a first-party, but it still says Meltano Community. 

@pnadolny13 think we should just update the label for `meltanolabs`?

PR #1328, opened 2023-05-18
Update variant list to use label

tayloramurphy:
#1327 

@pnadolny13 thoughts on this PR?


tayloramurphy:
@pnadolny13 ping here when you get a chance.

pnadolny13:
@tayloramurphy I'm good with renaming to remove the "Community" part. Its a little tricky though because something like tap-github is truly maintained by the community although its in meltanolabs. Maybe in the future we can differentiate them we wanted.

I open to the change but honestly can't decide how I feel about using the maintainer label as the variant name. On one hand its a nicer format but on the other it could be confusing to see the Simon Data variant listed on the hub but then you need to install it as the `radico` variant.

Screen Shot 2023-05-19 at 12 44 39 PM

Screen Shot 2023-05-19 at 12 44 43 PM

tayloramurphy:
didn't mean to close 🤦 

@pnadolny13 we could change the language from `Available Variants` to `Maintainer Variants` or `Contributor Variants`. We say in https://docs.meltano.com/concepts/plugins#variants "Alternative Implementations" which could work too. 

They'd have to click in to get the exact command to run still.

tayloramurphy:
@pnadolny13 updated to "Available Implementations"! If you'll approve we can merge.

Issue #993, opened 2022-11-02
bug: tap-greenhouse install fails


pnadolny13:
https://github.com/meltano/hub/pull/992#issuecomment-1300600587

It looks like it references a branch that no longer exists https://github.com/codyss/tap-greenhouse/blob/21a726bf1a783dd6a693cacebb33e19636ecbbd6/setup.py#L15.

```
Some possibly relevant errors from pip install:
error: pathspec 'master' did not match any file(s) known to git
error: subprocess-exited-with-error
Error installing tap-greenhouse from spec
'git+https://github.com/codyss/tap-greenhouse.git'.
```

edgarrmondragon:
@pnadolny13 The default branch was switched to `main`, but this package was never updated 🤦‍♂️

```diff
- "tap-kit @ git+https://github.com/Radico/tap-kit.git@master"
+ "tap-kit @ git+https://github.com/Radico/tap-kit.git@main"
```

What should we do in these cases? It's relatively easy to send [codyss/tap-greenhouse](https://github.com/codyss/tap-greenhouse) a PR, but not guarantees on it getting merged I guess.

pnadolny13:
Potentially use https://github.com/Pathlight/tap-greenhouse instead.

pnadolny13:
@rawwar just recently added their SDK based variant as the default! So the bug is less pressing although we should consider hiding codyss if it fails to install.

@rawwar in case you find it helpful, I guess theres https://github.com/Pathlight/tap-greenhouse that looks like it has pretty good stream coverage. Those schemas might help you accelerate adding more stream coverage to your variant!

rawwar:
For greenhouse, I will be including all streams mentioned in the documentation this weekend. Just an FYI

pnadolny13:
I opened https://github.com/meltano/hub/pull/1329 to hide the codyss variant that has install failures and chose not to add the Pathlight variant because the README still says its using the Gorgias API so its clearly a copy/pasted repo that was edited and maybe never finished? Its confusing either way so its not worth adding especially because rawwar's SDK variant looks like its making great progress with far more streams than the other variants.

PR #1329, opened 2023-05-18
hide codyss tap-greenhouse variant install failure

pnadolny13:
Closes https://github.com/meltano/hub/issues/993


tayloramurphy (approved):
@pnadolny13 what do you think about putting a comment in the YAML about why it was hidden and/or a link to the PR? Approving either way!

PR #1343, opened 2023-05-23
fix search

tayloramurphy:


pnadolny13 (approved):
@tayloramurphy  I cant speak for the JS specifically but the preview works as expected with all the searches I tried.

PR #1347, opened 2023-05-24
Add quality attribute - Part 2

pnadolny13:
Relates to https://github.com/meltano/hub/pull/1346


pnadolny13:
@tayloramurphy its probably easiest to iterate using the spreadsheet in https://github.com/pnadolny13/hub-utils/pull/45#issuecomment-1563233207 then once we're happy I'll propagate the change to this PR. 

I also updated this to only include and require quality for extractors/loaders instead of all plugins like I did before. Let me know if you were thinking differently though!

tayloramurphy (approved):

PR #1396, opened 2023-06-21
chore: refresh quality attributes

pnadolny13:
Closes https://github.com/meltano/hub/issues/1397

### DIFF
--- a/_data/meltano/extractors/tap-chargebee/airbyte.yml
+++ b/_data/meltano/extractors/tap-chargebee/airbyte.yml
@@ -17,7 +17,7 @@ name: tap-chargebee
 namespace: tap_airbyte
 next_steps: ''
 pip_url: git+https://github.com/MeltanoLabs/tap-airbyte-wrapper.git
-quality: silver
+quality: bronze
 repo: https://github.com/airbytehq/airbyte/tree/master/airbyte-integrations/connectors/source-chargebee
 settings:
 - description: Airbyte image to run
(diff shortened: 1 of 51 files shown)


_data/meltano/extractors/tap-chargebee/airbyte.yml
```diff
@@ -17,7 +17,7 @@ name: tap-chargebee
 namespace: tap_airbyte
 next_steps: ''
 pip_url: git+https://github.com/MeltanoLabs/tap-airbyte-wrapper.git
-quality: silver
```
tayloramurphy, line 20: @pnadolny13 what's contributing to the Airbyte downgrade? I would expect us to keep this at Silver given that Airbyte is actively maintaining these AFAICT.

pnadolny13: @tayloramurphy I couldnt find the thread where we talked about it but we set minimums of bronze for certain variants like matatika/transferwise/airbyte/singer-io. The tests are in https://github.com/pnadolny13/hub-utils/blob/8ad7f5c543c537ef7653a2243edf38f95a421b09/tests/test_meltano_utils.py#L478. These are getting downgraded because we dont have enough usage data. Its a rolling 3 month window so its possible that people were using it but then stopped.

tayloramurphy: @pnadolny13 makes sense then. Thanks for the reminder.

tayloramurphy (approved):

Issue #1389, opened 2023-06-16
Meltanolabs tap-postgres variant to default


tayloramurphy:
@pnadolny13 what's missing to make our variant the default?

PR #1399, opened 2023-06-26
meltanolabs tap and target postgres and snowflake to defaults

pnadolny13:
Closes https://github.com/meltano/hub/issues/1389

- tap and target postgres to default
- target snowflake to default and active maintenance
- tap-snowflake to default


tayloramurphy (approved):
Fantastic! @sbalnojan we need to publicize that we're making these changes and investing in the connectors.

PR #1403, opened 2023-06-27
Add tap-netsuite variant

sehnem:
## Checklist

- [x] Add/update the file in the appropriate folder (`/taps` or `/targets`). The name of the file should match the name of the tap. If there is already one, add a descriptor to the name such as `-search`.
- [x] Add/update the PNG logo image in `/assets/logos/<taps or targets>`. The image name must match the YAML file name.

## Reviewer Checklist


tayloramurphy:
Exciting to see! @pnadolny13 will you review and merge?

_data/meltano/extractors/tap-netsuite/sehnem.yml
```diff
@@ -0,0 +1,65 @@
+capabilities:
+- properties
+- discover
+- state
+description: Business Management Software
+domain_url: https://www.netsuite.com/
+keywords:
+- api
+- meltano_sdk
+label: NetSuite
```
pnadolny13, line 10:
```suggestion
label: NetSuite (SOAP)
```

pnadolny13 (approved):
@sehnem this is awesome!

Relative to the other variants on the hub like the gary james variant https://hub.meltano.com/extractors/tap-netsuite--gthesheep/, it looks like this uses the SOAP API while the other one uses REST. Is that true? I suggested adding SOAP to the label so people quickly see the difference.

What do you think?

sehnem:
@pnadolny13 I agree, just didn't want to change the current name, but it would make things much clearer, given that there are three versions of NetSuite API, SOAP, REST and SuiteQL, and there is also a difference on the schema of the streams, so we cannot just replace SOAP with REST.

sehnem:
I also set the quality to silver, I know that it is a new tap, but I think that overall it is more organized than the others and is using the SDK, although it need to be tested with more accounts. Let me know if you think it would be better to set it to bronze.

pnadolny13:
>@pnadolny13 I agree, just didn't want to change the current name, but it would make things much clearer, given that there are three versions of NetSuite API, SOAP, REST and SuiteQL, and there is also a difference on the schema of the streams, so we cannot just replace SOAP with REST.

This is a good point. Its a little weird to have a default variant in this case. I'm not very familiar with Netsuite, are the REST/SOAP/SuiteQL taps all attempting to extract somewhat similar data? Or is there totally different data coming from each different API.

It makes me wonder if we should split these out to their own standalone tap pages, we'd basically rename this tap on the hub to tap-netsuite-soap. We should merge this for now but if you think that makes sense I can open an issue to split them out eventually.

PR #1406, opened 2023-06-29
Adds extractors tap-mailchimp (acarter24)

pnadolny13:
@acarter24 do you have a good sense of the other variants? I assume you chose to rewrite it on the SDK because the other variants were lacking in some way. I'm curious what that reason is and anything else you know so we can decide what the best default is.


acarter24:
I think there was a couple reasons:

- using sdk to support `select` 
- Using the state for `report_email_activity` and unsubscribes to return less data from the api via the 'since' param (although I see this has now been partially delivered in the original variant https://github.com/singer-io/tap-mailchimp/pull/61)
- Avoid the batch endpoint, I didn't find it any quicker that making individual post requests for each campaign.

I wouldn't say its natually a better default that the existing, but there might be some folks who prefer an SDK version, and I would welcome a bit more feedback yet.

pnadolny13:
@acarter24 those are all valid reasons! We do prefer an SDK based variant as well because its more maintainable. It seems like yours likely has enough stream coverage and enough benefit to be made the new default.

pnadolny13:
@tayloramurphy this is good to go if you want to give it a final review.

tayloramurphy (approved):

PR #1446, opened 2023-07-23
feat: Add GtheSheep variant tap-contentful

GtheSheep:


pnadolny13:
@GtheSheep this is fantastic! What do you think about making this the default variant? I'm sure you looked at the existing variant so you'd probably know better but it looks like its very outdated and the settings arent even fully listed. I'd vote to make yours the default.

GtheSheep:
Hey @pnadolny13 - happy to! tbh I never got past installing the other one (it failed), and didn't look much into whether that was just a outdated package thing or a compatibility with Meltano thing, so reached straight for the SDK 😅

pnadolny13 (approved):

PR #1461, opened 2023-08-01
Add new config fields for Target Clickhouse

BTheunissen:
Updates settings to include `table_name` key for overriding loader table name.

### DIFF
--- a/_data/meltano/loaders/target-clickhouse/shaped-ai.yml
+++ b/_data/meltano/loaders/target-clickhouse/shaped-ai.yml
@@ -2,7 +2,7 @@ capabilities:
 - about
 - schema-flattening
 - stream-maps
-description: Clickhouse Database
+description: Loader for Clickhouse database.
 domain_url: https://www.clickhouse.com/
 executable: target-clickhouse
 keywords:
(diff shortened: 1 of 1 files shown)


_data/meltano/loaders/target-clickhouse/shaped-ai.yml
```diff
@@ -18,6 +18,14 @@ pip_url: git+https://github.com/shaped-ai/target-clickhouse.git
 quality: silver
 repo: https://github.com/shaped-ai/target-clickhouse
 settings:
+- description: The SQLAlchemy connection string for the ClickHouse database
+  kind: password
+  label: SQLAlchemy URL
+  name: sqlalchemy_url
+- description: The name of the table to write to. Defaults to stream name.
```
pnadolny13, line 25: @BTheunissen this is more of a target comment but doesnt this cause problems if you have multiple streams? If the two streams have different schemas and theyre both being forced into a single table, wont it though an error?

Either way thats a question for the target so from a hub perspective this looks good!

pnadolny13 (approved):

_data/meltano/loaders/target-clickhouse/shaped-ai.yml
BTheunissen: Yeah absolutely this mapping needs to be rethought at some point, honestly this is just a short term workaround I added on the target for my specific use-case. Do you have any examples you can think of where the Target config allows you to provide a mapping of entity stream to table? Otherwise I can take a look at refactoring this in the next run I take at it.

BTheunissen: I've added `hard_delete` and `add_record_metadata` into this PR also as I've updated the underlying SDK. Right now I've started to use the `table_name` config key in my deployment of the target so I'd like to merge as-is, but I am going to investigate doing this instead with stream mapping as a follow-up.

pnadolny13: @BTheunissen yeah 100%. We can always remove it in the future if you dont find it necessary.

PR #1486, opened 2023-08-21
Added matatika tap-capsulecrm

DanielPDWalker:
Add the Matatika variant `tap-capsulecrm` to the hub.


pnadolny13 (approved):
@DanielPDWalker thanks for the PR!

PR #1499, opened 2023-08-29
Include required start_date setting

amadrizwan:
The start_date is missing here. When we try to add start_date as an environment variable TAP_GOOGLE_SEARCH_CONSOLE_START_DATE="2019-01-01T00:00:00Z" it is not taken up. Currently it has to be defined in meltano.yaml to take effect.


pnadolny13:
@amadrizwan thanks for the PR!

PR #1508, opened 2023-09-05
Added matatika variant of tap-dbt-artifacts

DanielPDWalker:
Hey, we ended up forking and updating a `tap-dbt-artifacts` variant specifically for storing dbt test results from the `run_results.json` artifact. 

This is the variant of the tap we use in an upcoming article, example project and video so thought it would be best on the hub so anyone can access it easily.


pnadolny13:
@DanielPDWalker thanks for the PR!

I looked at the [code compare](https://github.com/prratek/tap-dbt-artifacts/compare/main...Matatika:tap-dbt-artifacts:main) because I was curious and I dont see anything thats materially different other than removing streams and asserting the file exists. Am I reading that right? What was the motivation to create a fork vs deselecting streams?

DanielPDWalker:
@pnadolny13 We unfortunately don't have support for selecting and deselecting streams in our UI at the moment, so we have to enable/disable the streams in the tap*.

Also at the moment it was a quick fix just get tap up and running for a customer demo around storing dbt test results, going forward we are likely going to have to make some more custom changes. Things we are thinking about:

- Supporting our chosen dbt versions, as these can produce different artifact formats, and the current logic doesn't support our "default" dbt version. 
- Adding support for the other artifacts as we need them (again of our "supported" dbt versions), so we can add more metadata to our test results for example
- *Might also look into allowing the selection of streams through a tap setting, eventually being removed when we add support for select i our UI.

pnadolny13:
@DanielPDWalker that makes sense - Do you have the option to install your tap as a custom plugin instead of from the hub? My thinking is that if its only useful for your custom use case then its not super valuable on the hub for others to use via meltano (where theyd just use select criteria) especially if you'll likely want to delist it after your UI has select capabilities. Although I also dont necessarily want to block it from being listed if its useful.

@edgarrmondragon @tayloramurphy any thoughts on this?

tayloramurphy:
@pnadolny13 I think I'm fine if we add it, but maybe we just hide it from the UI so it's at least accessible via the API for them?

pnadolny13:
>I think I'm fine if we add it, but maybe we just hide it from the UI so it's at least accessible via the API for them?

@DanielPDWalker what do you think about that idea? Seems reasonable to me

DanielPDWalker:
@pnadolny13 Yeah that would work great.

_data/meltano/extractors/tap-dbt-artifacts/matatika.yml
```diff
@@ -0,0 +1,23 @@
+capabilities:
+- catalog
+- discover
+description: dbt Artifact Files
+domain_url: https://docs.getdbt.com/reference/artifacts/dbt-artifacts#when-are-artifacts-produced
+keywords:
```
pnadolny13, line 6:
```suggestion
hidden: true
keywords:
```

pnadolny13 (approved):

PR #1512, opened 2023-09-08
Updates extractors tap-salesforce (SDK based - meltanolabs)

pnadolny13:
This points to the new `sdk` branch that contains the SDK version of the tap. `main` is still the legacy tap that is stable. This PR means that new users will get the SDK variant.

Based on the conversation in slack:

Screen Shot 2023-09-08 at 5 11 33 PM

### DIFF
--- a/_data/meltano/extractors/tap-salesforce/meltanolabs.yml
+++ b/_data/meltano/extractors/tap-salesforce/meltanolabs.yml
@@ -1,17 +1,23 @@
 capabilities:
-- properties
+- about
+- batch
+- catalog
 - discover
+- schema-flattening
 - state
+- stream-maps
 description: Customer-relationship management & customer success platform
-domain_url: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/intro_what_is_rest_api.htm
+domain_url: 
+  https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/intro_what_is_rest_api.htm
 keywords:
 - api
+- meltano_sdk
 label: Salesforce
 logo_url: /assets/logos/extractors/salesforce.png
 maintenance_status: active
 name: tap-salesforce
 namespace: tap_salesforce
-pip_url: git+https://github.com/meltanolabs/tap-salesforce.git@v1.5.1
+pip_url: git+https://github.com/meltanolabs/tap-salesforce.git@sdk
 quality: silver
 repo: https://github.com/meltanolabs/tap-salesforce
 select:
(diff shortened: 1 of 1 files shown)


edgarrmondragon (approved):

tayloramurphy (approved):

sicarul:
I'm debugging locally why because i don't still understand what happened, but since you uploaded this change, my local tap-salesforce started breaking, first because api_type is required, and now it doesn't want to parse my credentials. I had the version pinned like this:


```
plugins:
  extractors:
  - name: tap-salesforce
    variant: meltanolabs
    pip_url: git+https://github.com/meltanolabs/tap-salesforce.git@v1.5.1
    config:
      start_date: '2017-01-01T00:00:00Z'
      is_sandbox: false
      max_workers: 6
      select_fields_by_default: true
      api_type: REST
[...]
```

sicarul:
Replacing variant:meltanolabs
with this:
```
namespace: tap_salesforce
    capabilities:
      - properties
      - discover
      - state
    settings:
      - label: Username
        name: username
        placeholder: Ex. me@my-organization.com
      - kind: password
        label: Password
        name: password
        placeholder: Ex. *****************
      - documentation: https://hub.meltano.com/extractors/salesforce.html#salesforce-setup
        kind: password
        label: Security Token
        name: security_token
        placeholder: Ex. *****************
      - label: Client ID
        name: client_id
      - kind: password
        label: Client Secret
        name: client_secret
      - kind: password
        label: Refresh Token
        name: refresh_token
      - kind: date_iso8601
        label: Start Date
        name: start_date
      - kind: boolean
        label: Is Sandbox
        name: is_sandbox
        value: false
      - kind: options
        label: API Type
        name: api_type
        options:
        - label: REST
          value: REST
        - label: BULK
          value: BULK
        value: REST
      - kind: boolean
        label: Select Fields By Default
        name: select_fields_by_default
        value: true
      - kind: integer
        label: State Message Threshold
        name: state_message_threshold
        value: 1000
      - kind: integer
        label: Maximum number of threads to use
        name: max_workers
        value: 8
```
Which i got from this diff, fixed my issue

edgarrmondragon:
@sicarul Did you have a [lock file](https://docs.meltano.com/concepts/plugins/#lock-artifacts) for the plugin?

sicarul:
I did not, but it was not mandatory until Meltano 3.0.0 (Which i upgraded into today)

I'm just pointing this out in case other folks have the same problem

sicarul:
Can i also ask why are you guys switching to the sdk branch? It seems objectively worse (It has hardcoded streams instead of discovering each salesforce instance's schema)

tjvananne:
I experienced the same issue as @sicarul. The `git+https://github.com/meltanolabs/tap-salesforce.git@v1.5.1` version was working just fine for us until yesterday. The original error was that we weren't including the required `select_fields_by_default`. We've added that but are now experiencing the `Exception: Cannot create credentials from config.` issue.

We're using the username/password/security_token auth method.

If we want to keep using `v1.5.1` instead of this `sdk` version, is that possible? Maybe we can `@` a specific commit hash in the `pip_url`?

Thanks!

---

Config for more context:

```yaml
  - name: tap-salesforce
    variant: meltanolabs
    pip_url: git+https://github.com/meltanolabs/tap-salesforce.git@v1.5.1
    config:
      username: <our_user_email_address>
      start_date: '2021-01-01T00:00:00Z'
      api_type: BULK
      select_fields_by_default: true
      max_workers: 1
    select:
    - '*.*'
```

(We select all fields here and then explicitly select what SalesForce objects we want in the Airflow DAG that executes Meltano)

sicarul:
@tjvananne have you tried with the config i used?
https://github.com/meltano/hub/pull/1512#issuecomment-1712537512

tjvananne:
Thank you @sicarul - that does seem to be working

tjvananne:
It looks like if you remove `variant: meltanolabs` but keep `pip_url: git+https://github.com/meltanolabs/tap-salesforce.git@v1.5.1` then you get the non-sdk branch as well.

I'm guessing the `variant` setting overrides the `pip_url`? 

TIL

tayloramurphy:
Thanks for the feedback @sicarul and @tjvananne. We're going to revert the swap for now until custom objects are supported in the sdk-variant. Apologies for the trouble it caused, but glad there was a quick fix! cc @pnadolny13 @edgarrmondragon

PR #1528, opened 2023-09-21
Pin `tap-googleads` to `v0.4.0`

ReubenFrankel:
https://github.com/Matatika/tap-googleads/releases/tag/v0.4.0


edgarrmondragon (approved):
Thanks!

Issue #1526, opened 2023-09-19
Add Plugin: target-hdfs


joaopamaral:
### What is the plugin type?

loader

### Repository URL

https://github.com/Automattic/target-hdfs

### Description

Automattic

### Domain URL

https://arrow.apache.org/docs/python/generated/pyarrow.fs.HadoopFileSystem.html

PR #1536, opened 2023-10-02
Adds loaders target-hdfs (automattic)

pnadolny13:
Closes https://github.com/meltano/hub/issues/1526


edgarrmondragon (approved):
This is an interesting one!

PR #1575, opened 2023-11-07
Adding new config and fixing formatting for tap-rest-api-msdk.

s7clarke10:
Adding the following new parameters to support the following PR made to tap-rest-api-msdk : https://github.com/Widen/tap-rest-api-msdk/pull/39.

- backoff_type
- backoff_param
- backoff_time_extension
- store_raw_json_message

These parameters are used to support the following three features to delivery required functionality for a new API.

1. It contains enhanced logic in the tap discovery to cache credentials to avoid having to re-authenticate for each stream. The API was erroring due to too many OAuth Requests in a quick succession.
2. Optional backoff logic has been built in allowing tap-rest-api-msdk to respond to http retry-after messages. This is configurable to use either the header or message responses. There is also a new setting which adds some additional time because sometime the requested wait time is longer enough. This feature was built in to meet the API's backoff requirements.
3. Optionally provides the ability to store the whole raw message. The feature is enabled by setting the `store_raw_json_message` to `true`. This is useful if you wish to offload, the flattening functionality to the likes of dbt. Where I have used this feature I tend to select only the primary key, the replication key, and the `_sdc_raw_json field`.
The use case for this was a dynamic schema with optional fields/columns which were not available in every record. In this situation the schema discovery did not pick up every field leading to missing data, it was elected that storing the raw json record was important - to ensure all data is preserved.

Points 2-3 are optional, and so the tap's behaviour does not change. For Point 1, the caching will speed up discovery and ingestion as credentials are cached.

### DIFF
--- a/_data/meltano/extractors/tap-rest-api-msdk/widen.yml
+++ b/_data/meltano/extractors/tap-rest-api-msdk/widen.yml
@@ -58,7 +58,6 @@ settings:
   kind: string
   label: Backoff Type
   name: backoff_type
-  value: none
 - description: 'The header parameter to inspect for a backoff time.
     Optional: Defaults to `Retry-After`.'
   kind: string
@@ -168,17 +167,14 @@ settings:
   kind: string
   label: Pagination Limit Per Page Param
   name: pagination_limit_per_page_param
-  value: none
 - description: The name of the param that indicates the page/offset. Defaults to `None`.
   kind: string
   label: Pagination Next Page Param
   name: pagination_next_page_param
-  value: none
 - description: The size of each page in records. Defaults to `None`.
   kind: integer
   label: Pagination Page Size
   name: pagination_page_size
-  value: none
 - description: The pagination style to use for requests. Defaults to `default`.
   kind: string
   label: Pagination Request Style
@@ -193,7 +189,6 @@ settings:
   kind: integer
   label: Pagination Results Limit
   name: pagination_results_limit
-  value: none
 - description: The name of the param that indicates the total limit e.g. `total`, `count`.
     Defaults to `total`.
   kind: string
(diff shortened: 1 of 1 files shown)


_data/meltano/extractors/tap-rest-api-msdk/widen.yml
```diff
@@ -52,6 +52,25 @@ settings:
   kind: object
   label: AWS Credentials
   name: aws_credentials
+- description: 'The style of Backoff [message|header] applied to rate limited APIs.
+    Backoff times (seconds) come from response either the `message` or `header`.
+    Optional: Defaults to `None`.'
+  kind: string
+  label: Backoff Type
+  name: backoff_type
+  value: none
```
pnadolny13, line 61: @s7clarke10 similar to my comment in https://github.com/meltano/hub/pull/1576/files#r1385464176. Should these `value: none` settings be excluded?

s7clarke10: Quick question here. Should I even add value entries for defaults if tap-rest-api-msdk has defaults set in the tap. I'm starting to think that I should remove most of the value statements that I added (they are not really required).

pnadolny13: @s7clarke10 we're somewhat inconsistent on this across the hub. On one hand its nice for documenting the defaults on the plugin pages but on the other hand we could accidentally be overriding changes to the default in the tap if we dont also keep the hub definition up to date. So I generally feel fine including defaults for SDK based plugins and for non-SDK based plugins I only include them if we want to override the default in the tap (although its better to make actual code changes to the tap but sometimes thats not possible).

s7clarke10: Thanks for that feedback. On this basis I have just removed the none defaults.

s7clarke10:
I believe the review items are complete now and this is ready to be published. Thanks for the review.

pnadolny13 (approved):

PR #1576, opened 2023-11-07
Sybase - Pointing to the correct repo, and adding missing parameter.

s7clarke10:
Change Summary:

- Changing Sybase to point to the Production working repository.
- Adding missing `database` parameter
- Formatting

### DIFF
--- a/_data/meltano/extractors/tap-sybase/s7clarke10.yml
+++ b/_data/meltano/extractors/tap-sybase/s7clarke10.yml
@@ -84,7 +84,6 @@ settings:
   kind: string
   label: Tds Version
   name: tds_version
-  value: none
 - description: To emit a date as a date without a time component or time without an
     UTC offset. This is helpful to avoid time conversions or to just work with a date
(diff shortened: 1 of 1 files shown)


_data/meltano/extractors/tap-sybase/s7clarke10.yml
```diff
@@ -80,6 +84,7 @@ settings:
   kind: string
   label: Tds Version
   name: tds_version
+  value: none
```
pnadolny13, line 87:
```suggestion
```
@s7clarke10 it looks like the tap accepts this value or None. You should be able to just exclude the `value` attribute for this. If someone sets it then it will populate in the config.json but otherwise its excluded. `value` is for non-null defaults.

s7clarke10: Thank you for confirming this.

s7clarke10:
Thank you @pnadolny13 . I appreciate the review. I think the PR is good to go now.

pnadolny13 (approved):

PR #1583, opened 2023-11-13
feat: :sparkles: add ticketswap hubspot tap

tobiascadee:
A forked tap from the potloc variant containing the custom properties we need in the deals stream.

### DIFF
--- a/_data/meltano/extractors/tap-hubspot/ticketswap.yml
+++ b/_data/meltano/extractors/tap-hubspot/ticketswap.yml
@@ -1,0 +1,79 @@
+capabilities:
+- about
+- batch
+- catalog
+- discover
+- schema-flattening
+- state
+- stream-maps
+description: A Hubspot tap containing custom properties in the deals stream.
+domain_url: ticketswap.com
+executable: tap-hubspot
+keywords:
+- meltano_sdk
+label: Hubspot
+logo_url: /assets/logos/extractors/hubspot.png
+maintenance_status: active
+name: tap-hubspot
+namespace: tap_hubspot
+next_steps: ''
+pip_url: git+https://github.com/TicketSwap/tap-hubspot.git
+quality: unknown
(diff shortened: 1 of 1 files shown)


pnadolny13:
@tobiascadee thanks for the PR! I clicked into your repo and it doesnt look like theres any changes between yours and the potloc version, is the repo missing some expected commits?

>A forked tap from the potloc variant containing the custom properties we need in the deals stream.

Based on your description it seems like this might be custom to your implementation, is that right? Do you think this will be useful for others outside your organization? Usually we're happy to list any and all variants on the hub but in the case of hubspot I'm a little hesitant because theres already so many listed and the community is always struggling to figure out which variant to pick.

tobiascadee:
@pnadolny13 ah I see I haven't committed my changes yet. on the second point: I agree with your line of reasoning, its minor changes to the existing variant so it actually would not make that much sense to create a separate variant. what would be an alternative where I could refer to my custom repo but maybe use the existing variant? It would be nice to not have to have the code local when executing meltano.

pnadolny13:
>what would be an alternative where I could refer to my custom repo but maybe use the existing variant? It would be nice to not have to have the code local when executing meltano.

@tobiascadee The hub definitions are really just metadata mostly for helping meltano understand how to use the plugin (i.e. should it track incremental state, what settings are available, etc.) so using the potloc definition should be fine if you havent altered the metadata.

You should be able to install the potloc variant as normal then update your meltano.yml pip_url to reference your repo, after running a `meltano install extractor tap-hubspot --clean` you'll have your code along with the potloc meltano plugin metadata. Another less ideal option is to configure it as a custom plugin in your project https://docs.meltano.com/guide/plugin-management#custom-plugins but that requires you to define all the settings and capabilities yourself.

tobiascadee:
@pnadolny13 ah thanks! I was wondering if that would be possible but nice to know that's an alternative. I have some other sources where basically any other implementation works but I need some changes in the streams config.

PR #1595, opened 2023-11-17
Add extractor tap-launchdarkly

sicarul:


_data/meltano/extractors/tap-launchdarkly/pulumi.yml
```diff
@@ -0,0 +1,57 @@
+capabilities:
+- about
+- catalog
+- discover
+- schema-flattening
+- state
+- stream-maps
+description: This tap allows you to obtain Feature Flags from LaunchDarkly to be able
+  to detect which users/customers/etc have different feature flags activated.
```
edgarrmondragon, line 9:
```suggestion
description: Obtain Feature Flags from LaunchDarkly
```

_data/meltano/extractors/tap-launchdarkly/pulumi.yml
```diff
@@ -0,0 +1,56 @@
...
+pip_url: git+https://github.com/pulumi/tap-launchdarkly.git
+quality: bronze
+repo: https://github.com/pulumi/tap-launchdarkly
+settings:
+- description: The token to authenticate against the API service
+  kind: password
+  label: Auth Token
+  name: auth_token
+- description: Which environment to get activation data for (Usually a production
+    environment)
```
edgarrmondragon, line 28:
```suggestion
- description: Environment to get activation data for
```

_data/meltano/extractors/tap-launchdarkly/pulumi.yml
```diff
@@ -0,0 +1,55 @@
...
+settings:
+- description: The token to authenticate against the API service
+  kind: password
+  label: Auth Token
+  name: auth_token
+- description: Environment to get activation data for
+  kind: string
+  label: Environment
+  name: environment
+  value: production
```
edgarrmondragon, line 31: Is this value always available?

sicarul: It's a mandatory setting, the tap extracts one environment at a time

sicarul: Do you mean if it's possible to use LaunchDarkly without an environment? All the entities sit in an environment if that's the question
image

edgarrmondragon: Gotcha. My worry is that it's an arbitrary value. For example, a user may have named their production environment `prod` instead. Does that make sense? If so, we should probably remove the default value and instead use [`settings_group_validation`](https://docs.meltano.com/reference/plugin-definition-syntax/#settings_group_validation).

sicarul: Ohhhh yes you are right

sicarul: Addressed in https://github.com/meltano/hub/pull/1595/commits/4197f1ecfae6293f2bb92b336811321082b99ba7

edgarrmondragon (approved):
The docs build failed and then it succeeded after I hit retry, but Netlify hasn't notified GH it seems 🤷 .

Thanks @sicarul!

PR #1607, opened 2023-12-04
feat: :sparkles: add ticketswap tap-exact variant

tobiascadee:
add a ticketswap variant of the exact tap


_data/meltano/extractors/tap-exact/ticketswap.yml
```diff
@@ -0,0 +1,88 @@
...
+  label: Client Secret
+  name: client_secret
+- description: The bucket to contain the access and refresh token to authenticate against the API service
+  kind: string
+  label: Tokens S3 Bucket
+  name: tokens_s3_bucket
+- description: The S3 object key to contain the access and refresh token to authenticate against the API service
+  kind: string
+  label: Tokens S3 Key
+  name: tokens_s3_key
```
edgarrmondragon, line 35: Hi @tobiascadee! These seem very implementation-specific. Can you say a bit more about how you use them?

tobiascadee: I agree but did not know any other easy way to store the refresh token. the issue is that when a new access token is provided a new refresh token is also provided. so you need to store that token somewhere to be able to authenticate next time. this was my initial solution but maybe you have a better option? this works with any s3 location so its generic in that sense but I agree its AWS specific.

tobiascadee: I'm thinking that another option is to retrieve and set the latest tokens outside of the tap codebase, I would prefer the current solution but that might be an alternative. let me know what you think

tobiascadee: @edgarrmondragon if I need to remove this part please let me know, I want to put this tap into production :)

pnadolny13: > I want to put this tap into production :)

@tobiascadee I'll let Edgar respond to your question but I just wanted to add that you can already put this tap into production without needing it to be listed on the hub. See customer plugins for more details https://docs.meltano.com/guide/plugin-management/#custom-plugins. The hub is mostly for distributing plugins so other people can find and use them, if this is a custom implementation then it might not be suitable for the hub and if youre eager to get it into production then I'd recommend you add it as a custom plugin at least for now.

edgarrmondragon: @tobiascadee Can you add a short paragraph to the `settings_preamble` key describing the S3 requirement?

tobiascadee: I will! I didn't mean to say that this is a custom implementation tap, the s3 requirement is just to make sure the authentication works.

edgarrmondragon: Yeah, I agree and [this sort of feature](https://github.com/meltano/meltano/issues/2660) has been missing (and required) from Meltano for some time.

tobiascadee: @edgarrmondragon done!

edgarrmondragon (approved):
Thanks @tobiascadee!

Issue #1642, opened 2024-01-17
Unpin dbt-* dependencies


edgarrmondragon:
Currently `dbt-*` utility plugins pin dbt-core and dbt adapters to their minor versions. For example, dbt-snowflake:

https://github.com/meltano/hub/blob/70a05c82bdefdc310d1fdcabd37cc254181e2db1/_data/meltano/transformers/dbt-snowflake/dbt-labs.yml#L49

This goes against the current recommended practice of having unpinned Hub/lockfile dependencies and pinned `meltano.yml` dependencies and increasingly leads to users installing old versions of dbt and its adapters.

cc @pnadolny13

---

References:

* Slack: https://meltano-old.slack.com/archives/C01TCRBBJD7/p1705407328979619
* Linen: https://www.linen.dev/s/meltano/t/16312063/update-latest-dbt#f8ee732b-266c-4f0d-b574-9ed809d51373

pnadolny13:
@edgarrmondragon I dont remember why we chose to pin these versions. Maybe since we have 3 dependencies loosely defined we wanted to pin them and test before bumping the versions to make sure they are compatible 🤷 . Without pinning we could end up in a broken state where the latest versions dont work together.

With that said, I'm fine with unpinning them. Its recommended that users pin anyways.

edgarrmondragon:
> Without pinning we could end up in a broken state where the latest versions dont work together.

Yeah, I think dbt-core is released with constraints such you shouldn't be able to install incompatible adapters but pinning for now feels safer.

pnadolny13:
If thats true then I'm actually in favor of unpinning because we'll probably let these versions get stale before updating them again, so it might be better to update them now while its top of mind.

PR #1644, opened 2024-01-18
chore: Bump dbt utilities to use `dbt-core` 1.7.*

edgarrmondragon:
* Related: https://github.com/meltano/hub/issues/1642
* Did not update the legacy transformer plugins
* Updated all dbt adapter utility plugins to use `1.7.*` for both `dbt-core` and the corresponding adapter package.


pnadolny13 (approved):
@edgarrmondragon this looks good to me! Although I did note in https://github.com/meltano/hub/issues/1642#issuecomment-1898960572 that if its preferred to unpin then we can.

PR #1675, opened 2024-02-16
chore: Update tab-db2 plugin definition

danielptv:
Replace `kind: password` with `sensitive: true` and add newly introduced config options.


danielptv:
Looks like `sensitive: true` is not yet supported by the schema validation. Or have I missed something?

pnadolny13:
@danielptv thanks for the PR! `sensitive` is about to be added as part of https://github.com/meltano/hub/pull/1667#pullrequestreview-1885369164. That should allow theses tests to pass.

edgarrmondragon:
I don't think we should remove `kind: password` just yet :)

danielptv:
@edgarrmondragon should I keep both `kind: password` and `sensitive: true` or only `kind: password`?

danielptv:
I have added back `kind: password` so both are now present.

pnadolny13 (approved):

Issue #1545, opened 2023-10-10
feat: Improve settings UX


ReubenFrankel:
As a user of a plugin, I would like to be able to see the config required to get the tap working at a minimum first, before optional settings defined by the developer or inherited from the SDK.

---

For example, [`tap-shopify`](https://hub.meltano.com/extractors/tap-shopify/) was [initially added to the hub](https://github.com/meltano/hub/blob/be46311be9186f116ba84c71ba9c1029e47cdf2b/_data/meltano/extractors/tap-shopify/matatika.yml) with the following developer-defined settings:

- `access_token`
- `store`
- `admin_url`
- `start_date`
- `is_plus_account`

[Now](https://github.com/meltano/hub/blob/bfc7c70f8d68f466b02f3c3f72e45fd9be9148ca/_data/meltano/extractors/tap-shopify/matatika.yml), implicit SDK settings are merged with developer-defined settings:

- `access_token`
- `admin_url`
- `batch_config.encoding.compression`
- `batch_config.encoding.format`
- `batch_config.storage.prefix`
- `batch_config.storage.root`
- `flattening_enabled`
- `flattening_max_depth`
- `is_plus_account`
- `start_date`
- `store`
- `stream_map_config`
- `stream_maps`

The problem here is that `store`, for example, is a (conditionally) required setting and, due to alphabetical sorting, is hidden amongst settings that are not required and/or not defined by the developer originally. This makes it hard to see what a user needs to provide to get the tap working at a minimum.

The quickest fix here might be to stop sorting settings alphabetically in the plugin definition `.yml` files and display them in the order that they are defined, in an effort to group by importance and behaviour. It would also be good to group optional settings by whether they were explicitly defined by the developer or inherited from the SDK (given the plugin was built with the SDK) - maybe there could be an `Advanced` or `SDK-inherited` subsection? In the long-term, perhaps there needs to be a better implementation of `settings_group_validation`, in order to show the required (unconditionally and conditionally) settings first, followed by the unconditionally optional settings.

tayloramurphy:
@ReubenFrankel great idea. Is there any metadata on the settings coming from the SDK? We could likely add something to the SDK and key off that on the Hub.

ReubenFrankel:
@tayloramurphy I don't know anything about setting metadata from the SDK (unclear if that question was directed at me). I was thinking that you could infer these settings from `capabilities`, if the plugin was build with the SDK.

tayloramurphy:
[...]

PR #1703, opened 2024-03-07
Remove settings not listed in tap-rest-api-msdk readme

epapineau:
Several settings listed in the Meltano Hub docs do not actually exist for `tap-rest-api-msdk`. This PR reconciles the settings [listed in the tap's README](https://github.com/Widen/tap-rest-api-msdk/blob/main/README.md) with Meltano Hub docs.

### DIFF
--- a/_data/meltano/extractors/tap-rest-api-msdk/widen.yml
+++ b/_data/meltano/extractors/tap-rest-api-msdk/widen.yml
@@ -72,30 +72,6 @@ settings:
   kind: string
   label: Backoff Type
   name: backoff_type
-- description: Compression format to use for batch files.
-  kind: options
-  label: Batch Config Encoding Compression
-  name: batch_config.encoding.compression
-  options:
-  - label: Gzip
-    value: gzip
-  - label: None
-    value: none
-- description: Format to use for batch files.
-  kind: options
-  label: Batch Config Encoding Format
-  name: batch_config.encoding.format
-  options:
-  - label: Jsonl
-    value: jsonl
-- description: Prefix to use when writing batch files.
-  kind: string
-  label: Batch Config Storage Prefix
-  name: batch_config.storage.prefix
-- description: Root path to use when writing batch files.
-  kind: string
-  label: Batch Config Storage Root
-  name: batch_config.storage.root
 - description: Used for the Bearer Authentication method, which uses a token as part
     of the authorization header for authentication.
   kind: password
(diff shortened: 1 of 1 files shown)


ReubenFrankel:
@epapineau These settings exist by virtue of `tap-rest-api-msdk` having been built with the Meltano Singer SDK, which comes with some generic configurable features out-the-box (stream maps, schema flattening batch files). You'll find pretty much all extractors in the Hub built with the SDK list these additional settings that they don't explicitly define in source code, but still take effect. I do think this is a bit confusing from a UX standpoint, so I opened #1545 a while back.

epapineau:
@ReubenFrankel - Thank you for the context and linked issue, it definitely helps me better understand the Meltano doc. That being said - tap-rest-api-msdk flattens data by default and [AFAICT] doesn't use `flatten_enabled` anywhere in the repo. Would removing at least the flatten settings from this doc and leaving the remaining SDK settings make sense?

edgarrmondragon:
@epapineau I think the tap would at the same time have to be update to remove "flattening" from its set of "capabilities". See https://github.com/meltano/sdk/blob/bf996a4fc404f56c8688893b1abece1f1b878aa7/singer_sdk/tap_base.py#L183-L198

for the default implementation.

epapineau:
@edgarrmondragon is that the capabilities listed [here](https://github.com/Widen/tap-rest-api-msdk/blob/4f87c1adae00446388ebbe418c70b87c231856dc/meltano.yml#L8)?

edgarrmondragon:
> @edgarrmondragon is that the capabilities listed [here](https://github.com/Widen/tap-rest-api-msdk/blob/4f87c1adae00446388ebbe418c70b87c231856dc/meltano.yml#L8)?

@epapineau Nope, rather either this class

https://github.com/Widen/tap-rest-api-msdk/blob/4f87c1adae00446388ebbe418c70b87c231856dc/tap_rest_api_msdk/client.py#L12

or this one

https://github.com/Widen/tap-rest-api-msdk/blob/4f87c1adae00446388ebbe418c70b87c231856dc/tap_rest_api_msdk/streams.py#L40

should override the `capabilities` property to exclude flattening and others that don't apply for the tap:

```python
class RestApiStream(RESTStream):
    capabilities = [
        TapCapabilities.CATALOG,
        TapCapabilities.STATE,
        TapCapabilities.DISCOVER,
        PluginCapabilities.ABOUT,
        PluginCapabilities.STREAM_MAPS,
        PluginCapabilities.BATCH,
    ]
```

ReubenFrankel:
@epapineau This is now addressed by #1706 - see https://hub.meltano.com/extractors/tap-rest-api-msdk/

pnadolny13:
I'm going to close this PR since the changes in https://github.com/meltano/hub/pull/1706 resolve it.

epapineau:
Thank you so much! This is much more explicit - very helpful.

PR #1706, opened 2024-03-15
feat: separate default sdk settings for readability

pnadolny13:
Closes https://github.com/meltano/hub/issues/1545

The growth of default SDK settings is a good thing for consistency but makes the UX worse here because there are so many settings to understand. As suggested in the issue, I took a first iteration on separating defined settings and SDK default settings. I did it for the main list of setting and the detailed descriptions for each.

Implementation notes:
- I just chose to use a hardcoded list of settings that we know are from the SDK. I figured that we add new default settings so infrequently that we can just add them here as needed.
- I wasnt able to get the check for the `meltano_sdk` keyword to work for some reason. Someone else can give it a shot but keywords was always coming through empty for me
- I chose to collapse all SDK settings in both sections by default. The one downside I found is that if you click on one of the SDK settings in the first list it brings you down to the collapsed detailed SDK settings section but doesnt expand it for you. Not the worst but it would be ideal if it could automatically expand in that situation.

cc @ReubenFrankel

### DIFF
--- a/src/components/PluginSettingsSection.vue
+++ b/src/components/PluginSettingsSection.vue
@@ -9,12 +9,24 @@
         find the setting you're looking for, click on any setting name from the list:
       </p>
       <ul class="list-disc list-inside pl-4">
-        <li v-for="(setting, index) in settings" v-bind:key="index">
+        <li v-for="(setting, index) in definedSettings()" v-bind:key="index">
           <a :href="'#' + setting.name.replace(/\./g, '-') + '-setting'"
             ><code>{{ setting.name }}</code></a
           >
         </li>
       </ul>
+      <div v-if="sdkSettings().length > 0">
+        <details>
+          <summary class="text-xl pb-4 pt-4 font-bold font-hg">Expand To Show SDK Settings</summary>
+          <ul class="list-disc list-inside pl-4">
+            <li v-for="(setting, index) in sdkSettings()" v-bind:key="index">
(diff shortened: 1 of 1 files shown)


ReubenFrankel:
Awesome, that looks so much better!

Would it be possible to filter the SDK settings based on a plugin's defined `capabilities` too? For example, show `faker_config.locale`, `faker_config.seed`, `flattening_enabled`, `flattening_max_depth`, `stream_map_config` and `stream_maps` only when the `stream-maps` capability is set.

ReubenFrankel:
> > Would it be possible to filter the SDK settings based on a plugin's defined capabilities too?
> 
> @ReubenFrankel Do you have an example of this? I would think thats a bug if the plugin lists `stream_maps`, `stream_maps_config` etc. but the capability `stream-maps` isnt present. Like I think the solution in that case is to remove those settings from the definition completely since they arent supported, rather than just hiding them. Am I understanding this correctly?

@pnadolny13 Never mind - I forgot that those settings are coming from the YAML definition and are updated by the metadata refresh process (which I assume already takes into account `capabilities` when adding SDK settings). :sweat_smile:

edgarrmondragon (approved):
Nice!

pnadolny13:
>which I assume already takes into account capabilities when adding SDK settings

@ReubenFrankel the refresh process just runs `--about` and converts it into meltano definition yaml format so if the tap says it has those setting then theyre shown on the hub, if not it doesnt get added.

PR #1707, opened 2024-03-15
Auto-expand SDK settings detail section from link

ReubenFrankel:
From #1706:

> I chose to collapse all SDK settings in both sections by default. The one downside I found is that if you click on one of the SDK settings in the first list it brings you down to the collapsed detailed SDK settings section but doesnt expand it for you. Not the worst but it would be ideal if it could automatically expand in that situation.

cc @pnadolny13

### DIFF
--- a/src/components/PluginSettingsSection.vue
+++ b/src/components/PluginSettingsSection.vue
@@ -85,7 +85,13 @@
       </span>
 
       <div v-if="sdkSettings().length > 0">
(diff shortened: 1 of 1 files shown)


edgarrmondragon (approved):

pnadolny13:
@ReubenFrankel this is awesome! Thanks for fixing that.

PR #1709, opened 2024-03-15
Use presence of `meltano_sdk` keyword to display SDK settings

ReubenFrankel:
From #1706:

> I wasnt able to get the check for the meltano_sdk keyword to work for some reason. Someone else can give it a shot but keywords was always coming through empty for me

cc @pnadolny13 is this what  you had in mind?

### DIFF
--- a/src/components/PluginSettingsSection.vue
+++ b/src/components/PluginSettingsSection.vue
@@ -15,7 +15,7 @@
           >
         </li>
       </ul>
-      <div v-if="sdkSettings.length > 0 && $page.plugins.keywords.includes('meltano_sdk')">
+      <div v-if="sdkSettings.length > 0">
         <details>
           <summary class="text-xl pb-4 pt-4 font-bold font-hg">Expand To Show SDK Settings</summary>
           <ul class="list-disc list-inside pl-4">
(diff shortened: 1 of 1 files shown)


src/components/PluginSettingsSection.vue
```diff
@@ -9,18 +9,18 @@
...
-        <li v-for="(setting, index) in definedSettings()" v-bind:key="index">
-          <a :href="getSettingHref(setting)"
+        <li v-for="(setting, index) in definedSettings" v-bind:key="index">
+          <a :href="setting.href"
             ><code>{{ setting.name }}</code></a
           >
         </li>
       </ul>
-      <div v-if="sdkSettings().length > 0">
+      <div v-if="sdkSettings.length > 0 && $page.plugins.keywords.includes('meltano_sdk')">
```
pnadolny13, line 18: @ReubenFrankel this is great! In the case that someone implemented a setting that matches the default SDK setting name but isnt using the SDK, wouldnt they be ignored? Since its not meltano_sdk and definedSettings already filtered them out. Idk if this would actually ever happen though.

I think when I was trying to implement this I was trying to put the keyword check in the method/computed logic so all settings would be considered defined if theres no `meltano_sdk` keyword. If there is that keyword then they get split across the two defined/sdk settings lists.

Does that make sense? I'm not a front end guy, youre clearly way better at writing this logic than me 😅 , would that be hard to implement?

ReubenFrankel: > In the case that someone implemented a setting that matches the default SDK setting name but isnt using the SDK, wouldnt they be ignored? Since its not meltano_sdk and definedSettings already filtered them out. Idk if this would actually ever happen though.

Yeah, you're right. I tested by giving `tap-spreadsheets-anywhere` a `stream_maps` setting and it didn't show. I have a fix that involves moving keyword check into the computed logic, as you say! :sweat_smile:

https://github.com/meltano/hub/pull/1709/commits/baa841a6b584d18ac809f072ed4266ae3688fe74

pnadolny13 (approved):
@ReubenFrankel looks good to me! Thanks for the PR

PR #1721, opened 2024-03-25
fix: Add SDK keyword for all Airbyte wrapper plugins

ReubenFrankel:
The Airbyte wrapper plugin is built with the SDK and exposes SDK settings - it would therefore benefit from #1709.


pnadolny13 (approved):
@ReubenFrankel thanks for this! It is a little weird that its both built with airbyte and meltano but its actually correct and it adds back the default SDK setting for these.

ReubenFrankel:
@pnadolny13 Yeah, I had the same thoughts and arrived at the same conclusion that you did. :slightly_smiling_face: Thanks!

PR #1718, opened 2024-03-20
feat: Auto-select search

ReubenFrankel:
Closes #1619, addresses a related Vue.js warning and fixes some (I assume) intended but non-functional behaviour around search blur.

PR #1722, opened 2024-03-25
fix: Fragment links blocked by search autofocus

ReubenFrankel:
#1718 introduced an obscure bug in which the search autofocus blocks URL fragment links, such that a link to (for example) `#settings` doesn't auto-scroll to the section on load.

https://hub.meltano.com/extractors/tap-auth0#settings

This seems to only cause an issue for fast connections or when running locally.


edgarrmondragon (approved):

edgarrmondragon:
Thanks @ReubenFrankel!

PR #1724, opened 2024-03-27
Add some missing settings to tap-s3 (crowemi)

edgarrmondragon:


pnadolny13 (approved):
@edgarrmondragon I think this might be in the list of plugins ignored in the refresh automation because it has some complex settings that the utils cant handle yet.

PR #1725, opened 2024-03-27
Add missing `oauth_url` setting for `tap-zohosprints`

ReubenFrankel:
https://github.com/AutoIDM/tap-zohosprints/blob/52dc10d6912b74897ca5d62c55aa4ddd0b5fb333/tap_zohosprints/tap.py#L53-L55


ReubenFrankel:
My bad!

PR #1726, opened 2024-03-28
feat: :sparkles: Added the Ticketswap variant of target-redshift loader

tobiascadee:
the current redshift target is unmaintainted and not being updated (the last commit was 4 years ago). this is a new redshift target using the redshift connector dialect for sqlalchemy and using redshift connector to execute sql in the connector. the reason I am not just using the redshift dialect and default sql target methods is that this dialect will throw an error when creating or altering tables (so statements without any returning values).


edgarrmondragon (approved):
Thanks @tobiascadee!

PR #1727, opened 2024-03-31
Add workflow to test Meltano add/install for changed plugins

ReubenFrankel:


ReubenFrankel:
@pnadolny13 @edgarrmondragon How can I enable this new workflow for testing?

edgarrmondragon:
> @pnadolny13 @edgarrmondragon How can I enable this new workflow for testing?

I think it has to be in`main` before it starts running. Probably the best approach is to try it out on your fork.

ReubenFrankel:
OK, think this is ready for a review. You can check out example workflow runs from [this test PR](https://github.com/ReubenFrankel/hub/pull/1) and [this push](https://github.com/ReubenFrankel/hub/actions/runs/8562261959).

edgarrmondragon:
@ReubenFrankel should we update the title to reflect that this doesn't install the plugins?

ReubenFrankel:
@edgarrmondragon It does install: https://github.com/meltano/hub/blob/a6726beba9c3827fc3d853781557cdd31d34bc39/.github/workflows/test_meltano_add_install.yml#L60-L62

We can combine these into one step (i.e. `meltano add` without `--no-install`), if that is preferred.

edgarrmondragon:
> @edgarrmondragon It does install:
> 
> https://github.com/meltano/hub/blob/a6726beba9c3827fc3d853781557cdd31d34bc39/.github/workflows/test_meltano_add_install.yml#L60-L62
> 
> We can combine these into one step (i.e. `meltano add` without `--no-install`), if that is preferred.

Oh gotcha, I missed that!. Makes sense to have them as separate steps :)

edgarrmondragon (approved):

PR #1731, opened 2024-04-04
refactor: :fire: remove kind: password from sensitive parameters because of deprication

tobiascadee:
removing kind:password because of deprication


edgarrmondragon:
Hey @tobiascadee!

The warnings should be gone for these plugins once we ship https://github.com/meltano/meltano/pull/8486.

PR #1732, opened 2024-04-04
refactor: update target-redshift maintenance status and default

tobiascadee:
the transferwise redshift target is unmaintained, so updating that. would argue that the ticketswap variant should be default because of maintenance status and built with meltano sdk


edgarrmondragon (approved):
Seems fair

PR #1742, opened 2024-04-19
fix: Use correct `pip_url` for TicketSwap `target-redshift`

ReubenFrankel:
I've been helping someone on the Meltano Slack using `target-redshift` and noticed that the `pip_url` is pointing at [this pip package](https://pypi.org/project/target-redshift/), which (like the old default variant) hasn't been updated in over 4 years.

This means anyone trying to install `target-redshift` currently is probably not getting the version they think they're getting.


ReubenFrankel:
cc @tobiascadee

edgarrmondragon (approved):

PR #1754, opened 2024-05-01
fix: Code block content overflowing

ReubenFrankel:
For example: https://hub.meltano.com/extractors/tap-slack/#prereqs

Code blocks overflow their container
image

Horizontal scrollbar show for non-overflowing content
image

This PR address both these issues.


edgarrmondragon (approved):
Oh wow, thanks @ReubenFrankel!

PR #1757, opened 2024-05-07
remove tap-clickcast and tap-krow

aroder:


edgarrmondragon (approved):
I see the repos no longer exist. Thanks for updating here @aroder!

PR #1802, opened 2024-07-10
ci: Bump actions to use Node 20

edgarrmondragon:


edgarrmondragon:
Build passes on Netlify with Node 20 after setting the `NODE_OPTIONS=--openssl-legacy-provider` env var.

https://github.com/gridsome/gridsome/issues/1694#issuecomment-2081375505

PR #1824, opened 2024-08-27
refactor: :fire: remove `dialect+driver` and `sqlalchemy_url` from settings

tobiascadee:
this target uses `redshift_connector` instead of sqlalchemy so removing those settings.

### DIFF
--- a/_data/meltano/loaders/target-redshift/ticketswap.yml
+++ b/_data/meltano/loaders/target-redshift/ticketswap.yml
@@ -40,13 +40,13 @@ settings:
   kind: integer
   label: Batch Size Rows
   name: batch_size_rows
-- description: Redshift cluster identifier. Note if sqlalchemy_url is set or enable_iam_authentication
+- description: Redshift cluster identifier. Note if enable_iam_authentication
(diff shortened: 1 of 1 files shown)


_data/meltano/loaders/target-redshift/ticketswap.yml
edgarrmondragon, line 1: @tobiascadee there are other references to `sqlalchemy_url` in some of the settings, should we update them too?

edgarrmondragon (changes requested):

_data/meltano/loaders/target-redshift/ticketswap.yml
tobiascadee: yes, will do! do you also want me to incorporate [your changes ](https://github.com/meltano/hub/pull/1819) in here as well?

edgarrmondragon: It's fine to keep separate PRs

edgarrmondragon (approved):
Thanks @tobiascadee!

PR #1837, opened 2024-09-11
Updated the `tables` option description of tap-s3-csv (Steve Clarke variant)

s7clarke10:
### Changes

**Depreciated**
- string_overrides
- guess_types

**New**

**datatype_overrides**: A object / dictionary of header names in the file and any override datatype other than string you with to set as the datatype. Example config - ``` "datatype_overrides":{"administration_number":"integer","percentage":"number","grade":"integer"}```.

### DIFF
--- a/_data/meltano/extractors/tap-s3-csv/s7clarke10.yml
+++ b/_data/meltano/extractors/tap-s3-csv/s7clarke10.yml
@@ -68,27 +68,35 @@ settings:
   name: table_suffix
 - description: "An array that consists of one or more objects that describe how to
     find files and emit records. Required - `table_name` and `search_pattern`. Optional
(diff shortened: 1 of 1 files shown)


s7clarke10:
This is a healthy update to the pipelinewise variant of tap-s3-csv which hasn't been updated for a long-time. This has been patched and updated to ensure the ci pipelines work for PR's.

Recommend this becomes the default instead of pipelinewise as there is no support for pipelinewise anymore.

edgarrmondragon (approved):

edgarrmondragon:
> This is a healthy update to the pipelinewise variant of tap-s3-csv which hasn't been updated for a long-time. This has been patched and updated to ensure the ci pipelines work for PR's.
> 
> Recommend this becomes the default instead of pipelinewise as there is no support for pipelinewise anymore.

https://github.com/meltano/hub/pull/1838

PR #1840, opened 2024-09-13
chore: Fill in missing metadata

edgarrmondragon:


edgarrmondragon:
I opened PRs to fix installation of the failing extractors, so they'll hopefully be installable again one day, otherwise installing from the forks should work.

PR #1848, opened 2024-09-19
Add the TicketSwap variant of `tap-apple-search-ads`

tobiascadee:
this is meltano sdk variant of `tap-apple-search-ads`. A lot cleaner and less code than the alternative.


edgarrmondragon (approved):
Thanks @tobiascadee!

PR #1780, opened 2024-06-10
chore: [Hub Bot] Refresh metadata 2024-06-10

MeltyBot:
Updates Plugin Definitions

PR #1849, opened 2024-09-19
chore: split metadata extract across several workflows

pnadolny13:
Related to https://github.com/meltano/hub/pull/1780#issuecomment-2173847297

Its still nice to use github actions for the fan out since we can iterate through python versions until one works. We can probably do that in python but we'd have to sort of rearchitect the actions probably to pass around a list of plugins to attempt and they get removed if successful so we can still iterate through python versions. To save work for now I'm attempting to leave it how it is and just split across multiple workflows.

If this doesnt work for extractors I can further split it to chunk out 100 at a time or something like that.

Unfortunately I need to merge this into main before I can test them, I will follow up afterwards if I need to fix anything.

### DIFF
--- a/.github/workflows/metadata-extract-airbyte.yml
+++ b/.github/workflows/metadata-extract-airbyte.yml
@@ -1,0 +1,89 @@
+name: Metadata Extract
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: 'The environment to run the workflow in'
+        required: true
+        type: choice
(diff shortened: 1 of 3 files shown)


pnadolny13:
> Unfortunately I need to merge this into main before I can test them, I will follow up afterwards if I need to fix anything.

@edgarrmondragon because of this I'll power merge this as admin so I can continue dev/testing.

PR #1858, opened 2024-09-30
Update OAuth instructions to use curl to get refresh token

thena-seer-sfg:
# Summary

The current instructions are incorrect. The code received in the redirect url is not a refresh token but a short-term authorization code, and the tap fails when it tries to use it as one. In order to get a refresh token, there is a code exchange step. The easiest way to do this is using curl, so I've added this step to the instructions.

I have also opened this PR in the source repo, where it seems these instructions were copied from: https://github.com/singer-io/tap-harvest-forecast/pull/28

# Test plan
Follow the updated instructions when setting up the extractor in meltano. Verify `meltano config tap-harvest-forecast test` confirms the validity of the configuration.


edgarrmondragon (approved):
Thanks @thena-seer-sfg!

PR #1859, opened 2024-09-30
Fix formatting for codeblock

thena-seer-sfg:
Oops, after verifying the deployment of https://github.com/meltano/hub/pull/1858, I see that this change needs to be made, because the first line is being cut off, as seen in this  screenshot:

Screenshot 2024-09-30 at 10 35 38 AM


edgarrmondragon (approved):
Ah, thanks again!

I've confirmed the code block looks right in https://deploy-preview-1859--meltano-hub.netlify.app/extractors/tap-harvest-forecast.

Let me know if it looks as you expect it too 🙂

tayloramurphy (approved):

thena-seer-sfg:
Looks good to me!

PR #1861, opened 2024-10-03
fix: updated two logos

jenniferabowd:
These logos do not scale down well and the text is not required.


edgarrmondragon (approved):
LGTM!

PR #1872, opened 2024-10-22
Update `tap-googleads`

ReubenFrankel:
Bump `tap-googleads` to the [latest version](https://github.com/Matatika/tap-googleads/releases/tag/v0.8.0) and add new settings.

Issue #1418, opened 2023-07-07
Compatibility with upcoming deprecation of plugin setting `kind: hidden` and `kind: password`


ReubenFrankel:
When https://github.com/meltano/meltano/pull/7733 and https://github.com/meltano/meltano/pull/7874 are released in a future Meltano version, `kind: hidden` and `kind: password` will become deprecated and show a warning message if existing setting definitions use them.

![image](https://github.com/meltano/meltano/assets/60552974/eea932a1-db32-4d71-ba1d-7ed7f1acb431)

### Problem
The deprecation message will show until the plugin definition is updated (i.e. plugin `.lock` file no longer contains deprecated `kind`s) on the Hub and `meltano lock --update` is run. Since all versions of Meltano reference the same Hub instance, removing the deprecated `kind`s for all plugin definitions would potentially break all but the latest versions of Meltano if a user was to ever update their plugin `.lock` files (they don't know anything about the new `hidden: true` and `sensitive: true` setting properties, unless some of they are back-ported).

#### `kind: hidden`

```yml
- name: oauth_token_proxy_url
  kind: hidden
  value: https;//proxy.host/api/oauth/token
```

changes to

```yml
- name: oauth_token_proxy_url
  hidden: true
  value: https;//proxy.host/api/oauth/token
```

#### `kind: password`

```yml
- name: api_key
  kind: password
```

changes to

```yml
- name: api_key
  sensitive: true
```

#### `kind` as actual data type
Consider providing actual expected data types as `kind` now that this is possible, if applicable for a plugin. I imagine that the majority of plugins will not require this, but consider the following setting:

```yml
- name: poll_interval_ms
  kind: hidden
  value: 5000
```

should probably change to

```yml
- name: poll_interval_ms
  kind: integer
  hidden: true
  value: 5000
```

pnadolny13:
@ReubenFrankel thanks for the write up!

How would we do this without breaking existing user's projects? We dont have yaml schema versioning of plugins on the hub so I'm not sure how we'd be able to transition to this new schema. It seems like we should be able to add the new keys to the definitions but we couldnt change the kind otherwise any non-latest meltano versions would stop treating that setting as sensitive. When would we be safe to remove them so the deprecation warnings go away?

ReubenFrankel:
[...]

PR #1873, opened 2024-10-22
feat: Do not display hidden settings

ReubenFrankel:
Follow-on from #1872 and related to #1418

### DIFF
--- a/src/components/PluginSettingsSection.vue
+++ b/src/components/PluginSettingsSection.vue
@@ -155,10 +155,12 @@ export default {
   },
   computed: {
     $settingsWithHref() {
-      return this.settings.map((setting) => ({
-        ...setting,
-        href: `#${setting.name.replace(/\./g, "-")}-setting`,
-      }));
+      return this.settings
+        .filter((setting) => !setting.hidden && setting.kind !== "hidden")
+        .map((setting) => ({
+          ...setting,
+          href: `#${setting.name.replace(/\./g, "-")}-setting`,
+        }));
     },
     $isSdkPlugin() {
       return this.$page.plugins.keywords.includes("meltano_sdk");
--- a/src/templates/Plugins.vue
+++ b/src/templates/Plugins.vue
@@ -542,6 +542,7 @@ query Plugins($path: String!, $name: String!) {
       label
       description
       description_rendered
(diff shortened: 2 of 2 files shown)


ReubenFrankel:
I believe `tap-googleads` is currently the only plugin with any hidden settings: https://github.com/meltano/hub/blob/fb5940bebe1cdfb95a0c78e283fa901b622ba9da/_data/meltano/extractors/tap-googleads/matatika.yml#L54

edgarrmondragon:
I'm having second thought on the semantics of `hidden`.

@ReubenFrankel do you feel like it's more intended for UIs that display (and set) Meltano plugin config? For example, for settings that may not be user-friendly.

ReubenFrankel:
> I'm having second thought on the semantics of `hidden`.
> 
> @ReubenFrankel do you feel like it's more intended for UIs that display (and set) Meltano plugin config? For example, for settings that may not be user-friendly.
> 
> So, I wonder if we need a `deprecated` flag to better reflect situations like this.

We use `hidden` for that exact purpose in our UI, where some settings are hidden deliberately as they should be configured by the system only. I understand where you are coming from though, it's just odd that `hidden` isn't (I don't think) used anywhere in the CLI or Hub currently - I guess it was a hang-over from the old Meltano UI?

Totally fine by me to drop this, especially since it's just `tap-googleads` that this would "benefit". Until more plugins are actively deprecating settings, I don't think the flag is super necessary.

edgarrmondragon:
@ReubenFrankel let's just keep the deprecation notice in the setting description. We can look at adding a `deprecated` to Meltano and the Hub plugin schema flag later on.

edgarrmondragon:
> I guess it was a hang-over from the old Meltano UI?

Yeah, but it also doesn't cost us much keeping it around, and is still useful for anyone trying to implementing any sort of settings UI for Meltano 🙂

ReubenFrankel:
> > I guess it was a hang-over from the old Meltano UI?
> 
> Yeah, but it also doesn't cost us much keeping it around, and is still useful for anyone trying to implementing any sort of settings UI for Meltano 🙂 

Fair enough, definitely not proposing removing it, especially since we use it 😅

PR #1877, opened 2024-10-24
chore: Use uv to run Meltano add and install test

edgarrmondragon:


.github/workflows/test_meltano_add_install.yml
```diff
@@ -59,4 +69,4 @@ jobs:
 
       - name: Test install
         working-directory: test-project
-        run: meltano install --force
```
edgarrmondragon, line 62: https://github.com/meltano/meltano/pull/8852

pnadolny13 (approved):

PR #1945, opened 2025-02-10
fix: Use plugin executable name to dispatch commands

edgarrmondragon:


edgarrmondragon:
/test-plugin name=meltano-map-transfomer pip-url=meltano-map-transform executable=meltano-map-transform

PR #1948, opened 2025-02-13
fix: singer-io/tap-fixerio doesn't support `--discover`

edgarrmondragon:
https://meltano.slack.com/archives/C069CQNHDNF/p1739436908773209

### DIFF
--- a/_data/meltano/extractors/tap-fixerio/singer-io.yml
+++ b/_data/meltano/extractors/tap-fixerio/singer-io.yml
@@ -1,5 +1,4 @@
 capabilities:
-- discover
 - state
 description: Foreign Exchnage Rates API
 domain_url: https://fixer.io/documentation


edgarrmondragon:
Ok, this is weird. The tap's source doesn't have `--discover` and `--properties`, but pulling it from PyPI does:

```console
$ uvx -p 3.9 tap-fixerio --help
Installed 13 packages in 10ms
usage: tap-fixerio [-h] -c CONFIG [-s STATE] [-p PROPERTIES] [--catalog CATALOG] [-d]

optional arguments:
  -h, --help            show this help message and exit
  -c CONFIG, --config CONFIG
                        Config file
  -s STATE, --state STATE
                        State file
  -p PROPERTIES, --properties PROPERTIES
                        Property selections: DEPRECATED, Please use --catalog instead
  --catalog CATALOG     Catalog file
  -d, --discover        Do schema discovery
```

edgarrmondragon:
Oh, it's because the tap is using `singer.utils.parse_args`, which all the Singer standard CLI options: https://github.com/singer-io/singer-python/blob/ae50276b7055248273d0458b85d891773f7d4597/singer/utils.py#L126.

The tap doesn't use anything other than `--config/-c` and `--state/-s` though, so this PR is correct.

edgarrmondragon:
https://github.com/singer-io/tap-fixerio/pull/15.

PR #1977, opened 2025-03-24
ci: Use `UV_BUILD_CONSTRAINT` to use a working version of `setuptools`

edgarrmondragon:
Waiting for upstream's decision on whether to roll back so we can decide ourselves if we need this workaround or not.

## Related

- https://github.com/pypa/setuptools/pull/4870
- https://github.com/pypa/setuptools/issues/4910


edgarrmondragon:
No longer needed

PR #2017, opened 2025-05-15
Update target-mysql - thkwag - Username is User

sicarul:
The property to set the username in target-mysql is user, not username.


edgarrmondragon (approved):
Thanks @sicarul!

PR #2026, opened 2025-05-29
fix: `tap-msaccess` definitions

ReubenFrankel:
Some MSAccess taps were broken or missing settings

### DIFF
--- a/.github/workflows/test-plugin-command.yml
+++ b/.github/workflows/test-plugin-command.yml
@@ -120,35 +120,35 @@ jobs:
       id: plugin-help
       working-directory: ${{ github.workspace }}/project
       run: |
-        meltano invoke ${{ github.event.inputs.name }} --help 2>&1 | tee help-output.txt
+        meltano invoke ${{ github.event.inputs.name }} --help | tee help-output.txt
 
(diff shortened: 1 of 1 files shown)


edgarrmondragon (approved):

ReubenFrankel:
All good! I think the Singer SDK version is wrong though.

edgarrmondragon:
> All good! ~I think the Singer SDK version is wrong though.~ Oops, never mind! They're pinned to an older release ([`v0.1.0`](https://github.com/Matatika/tap-msaccess/releases/tag/v0.1.0)).

Oh then the schedule settings auto-update here will probably try to revert these changes. I can just ignore them though, knowing this is the reason.

ReubenFrankel:
> > All good! ~I think the Singer SDK version is wrong though.~ Oops, never mind! They're pinned to an older release ([`v0.1.0`](https://github.com/Matatika/tap-msaccess/releases/tag/v0.1.0)).
> 
> Oh then the schedule settings auto-update here will probably try to revert these changes. I can just ignore them though, knowing this is the reason.

Auto-update of what, sorry?

PR #2219, opened 2026-02-23
Add common_topic setting to target-kafka (think41)

rishipradeep-think41:
Added a common topic setting to allow users to easily specify a common topic to push all streams to


edgarrmondragon:
Thanks @rishipradeep-think41!

edgarrmondragon (approved):

PR #2311, opened 2026-08-11
feat(loaders): add target-s3 (plinxore variant)

plinxore:
## Summary

Adds the `plinxore` variant of `target-s3`, a fork of [`crowemi/target-s3`](https://github.com/crowemi/target-s3) matured for production use:

- Real gzip compression (the original wrote plaintext behind a misleading `.gz` extension)
- Corrected JSONL datetime serialization
- Configurable handling of source dates the SDK can't parse (e.g. legacy MySQL/MyISAM zero-dates)
- A collision-safe per-batch filename scheme (monotonic counter + UUID) so concurrent or repeated runs can never silently overwrite each other's files
- A Parquet writer with a schema kept stable across batches (derived once from the Singer SCHEMA rather than re-inferred per batch, so multi-file glob reads never hit a schema mismatch) and exact decimal precision for DECIMAL-shaped columns (`decimal128`, not `float64`)

Package: [`plinxore-target-s3`](https://pypi.org/project/plinxore-target-s3/) on PyPI (currently `0.1.2`).
Repo: https://github.com/plinxore/target-s3

Settings were scraped directly from the published package's `--about --format json` output via `hub-utils`, and validated with `plugin_schema_validate.py`, `yaml_check_names.py`, and `maintainers_validate.py`.

## Test plan

- [x] `plugin_schema_validate.py` passes for all plugin definitions
- [x] `yaml_check_names.py` passes
- [x] `maintainers_validate.py` passes
- [x] Settings scraped from a real `pip install plinxore-target-s3==0.1.2` in a fresh venv, confirming `format_type` includes `parquet` and `format.format_parquet.get_schema_from_tap` reflects the real default/description

### DIFF
--- a/_data/meltano/loaders/target-s3/plinxore.yml
+++ b/_data/meltano/loaders/target-s3/plinxore.yml
@@ -295,5 +295,6 @@ supported_python_versions:
 - '3.9'
 - '3.10'
 - '3.11'
+- '3.12'
 usage: ''
 variant: plinxore
--- a/_data/meltano/loaders/target-s3/plinxore.yml
+++ b/_data/meltano/loaders/target-s3/plinxore.yml
@@ -8,11 +8,11 @@ description: "S3-compatible object storage target (JSON/JSONL/CSV/Parquet). This
   JSONL datetime serialization, configurable handling of source dates the SDK can't
   parse (e.g. legacy MySQL zero-dates), a collision-safe per-batch filename scheme
   (a monotonic counter plus a UUID) so concurrent or repeated runs can never silently
-  overwrite each other's files, and a Parquet writer with a stable schema across
-  batches (derived once from the Singer SCHEMA rather than re-inferred per batch,
-  so files from the same stream can always be read together) and exact decimal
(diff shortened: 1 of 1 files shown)


plinxore:
The failing "Dispatch tests" check doesn't look related to this PR's content — the plugin definition passes `plugin_schema_validate.py`, `yaml_check_names.py`, and `maintainers_validate.py` locally, and "Get changed files" now passes.

Looking at the job logs, the failure happens when the workflow tries to trigger the internal test runner:

```
POST /repos/meltano/hub/actions/workflows/test-plugin-command.yml/dispatches
< 401
##[error]Bad credentials - https://docs.github.com/rest
```

For comparison, #2307 (a similar new-plugin PR, merged 2026-07-29) used the same dispatch mechanism and passed without issue. Something changed between then and now — likely the token/secret this workflow uses to dispatch `test-plugin-command.yml` has expired or been rotated. This would affect any PR touching a plugin definition right now, not just this one.

Flagging in case it's useful — happy to re-trigger once it's sorted out.

PR #2319, opened 2026-09-11
Add tap-fxmacrodata

roberttidball:
Adds an SDK-based extractor for [FXMacroData](https://fxmacrodata.com/?utm_source=meltano&utm_medium=referral&utm_campaign=open_source_integrations&utm_content=pr): official macroeconomic, FX and release-calendar data for 18 currencies, aggregated from central banks and national statistics agencies.

Repo: https://github.com/fxmacrodata/tap-fxmacrodata

### Streams

| Stream | Grain | Replication |
| --- | --- | --- |
| `announcements` | one record per release of one indicator | incremental on `date` |
| `release_calendar` | scheduled and recent releases | full table |
| `data_catalogue` | indicators a currency publishes | full table |
| `forex_rates` | daily reference rate for a pair | incremental on `date` |

Every `announcements` record carries `announcement_datetime`, the instant the figure was published. Most macro sources give a value and a reference date, which quietly invites lookahead — the figure for March was not knowable in March — so keeping the publication instant is the point of the stream.

`announcements` and `forex_rates` are partitioned per series, so state is held per currency/indicator and per pair rather than for the stream as a whole.

### Trying it without credentials

USD macro data is public, so the tap runs with no API key:

```json
{
  "currencies": ["USD"],
  "indicators": ["inflation", "policy_rate"],
  "start_date": "2020-01-01T00:00:00Z"
}
```

A key widens it to the other seventeen currencies, the full history and FX, and is sent as a header. A currency the key does not cover logs a warning and yields no rows rather than failing the run, so a mixed currency list stays usable.

### How this definition was produced

`hub-utils add --repo-url https://github.com/fxmacrodata/tap-fxmacrodata` installed the tap from the `pip_url` correctly, then failed at the settings scrape with `[WinError 2] The system cannot find the file specified` — it could not locate the installed executable on Windows. The definition is therefore written from the tap's own `--about --format=json` output, using the same transformation, and validated against `schemas/plugin_definitions/extractors.schema.json`.

Happy to regenerate it with the tool if a maintainer on Linux would rather have it produced that way.

Also updates `_data/default_variants.yml` and `_data/maintainers.yml`, both placed alphabetically.

### DIFF
--- a/_data/default_variants.yml
+++ b/_data/default_variants.yml
@@ -214,6 +214,7 @@ extractors:
   tap-frontapp: singer-io
   tap-fulfil: fulfilio
   tap-fullstory: singer-io
+  tap-fxmacrodata: fxmacrodata
   tap-ga4: meltanolabs
   tap-gainsightpx: widen
   tap-geekbot: edgarrmondragon
--- a/_data/maintainers.yml
+++ b/_data/maintainers.yml
@@ -490,6 +490,10 @@ full360:
   label: full360
   name: full360
   url: https://github.com/full360
+fxmacrodata:
+  label: FXMacroData
+  name: fxmacrodata
+  url: https://github.com/fxmacrodata
 gabilew:
   label: gabilew
   name: gabilew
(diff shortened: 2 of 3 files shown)


edgarrmondragon:
Thanks!

